PR #3, opened 2018-11-28
This should be initial CI config

s-rubenstein:
This is copied over with a lot of the workbench specific stuff removed, and the directories updated. There are a bunch of TODO's in here also, which will need to get done as we break things out.


.circleci/config.yml
```diff
@@ -0,0 +1,232 @@
...
+          - api-cache-
+      - run:
+          working_directory: ~/data-browser
+          command: ci/activate_creds.sh api/circle-sa-key.json
+      - deploy:
+          name: Deploy to App Engine
+          working_directory: ~/data-browser/api
+          command: |
+            ./project.rb deploy \
+              --project all-of-us-workbench-test \
```
calbach: aou-db-test

.circleci/config.yml
```diff
@@ -0,0 +1,232 @@
...
+          - api-cache-{{ checksum "~/data-browser/api/build.gradle" }}
+          - api-cache-
+      - run:
+          working_directory: ~/data-browser
+          command: ci/activate_creds.sh api/circle-sa-key.json
+      - deploy:
+          name: Deploy to App Engine
+          working_directory: ~/data-browser/api
+          command: |
+            ./project.rb deploy \
```
calbach: deploy-public-api

.circleci/config.yml
```diff
@@ -0,0 +1,232 @@
...
+      - restore_cache:
+          keys:
+          - api-cache-{{ checksum "~/data-browser/api/build.gradle" }}
+          - api-cache-
+      - run:
+          working_directory: ~/data-browser
+          command: ci/activate_creds.sh api/circle-sa-key.json
+      - deploy:
+          name: Deploy to App Engine
+          working_directory: ~/data-browser/api
```
calbach: You'll probably need to run this a bunch and iterate, but at least one issue here:  api --> public-api

.circleci/config.yml
```diff
@@ -0,0 +1,232 @@
...
+          command: |
+            ./project.rb deploy \
+              --project all-of-us-workbench-test \
+              --account circle-deploy-account@all-of-us-workbench-test.iam.gserviceaccount.com \
```
calbach: Could you create a circle-deploy-account under `aou-db-test`? Just added your pmi-ops as an owner

.circleci/config.yml
```diff
@@ -0,0 +1,232 @@
...
+      # Use the ui-build-test workspace here to avoid redoing the setup.
+      - attach_workspace:
+          at: .
+      - deploy:
+          name: Deploy to App Engine
+          working_directory: ~/data-browser/public-ui
+          command: |
+            ../ci/activate_creds.sh circle-sa-key.json
+            ./project.rb deploy-ui \
+              --project all-of-us-workbench-test \
```
calbach: Please rm all references to this project

ci/Dockerfile.circle_build
```diff
@@ -0,0 +1,41 @@
+# To build and deploy, from this directory:
```
calbach: Possible this could be omitted for now, and we continue using the workbench image. The circle config above still uses the older one

ci/upload_cloud_creds.sh
```diff
@@ -0,0 +1,37 @@
...
+        https://circleci.com/api/v1.1/project/$1/$2/$3/envvar?circle-token=$CI_TOKEN \
+        -d "{\"name\": \"$4\", \"value\": \"$5\"}"
+}
+
+GCLOUD_CREDENTIALS_KEY=$(openssl rand -base64 32)
+GCLOUD_CREDENTIALS=$(openssl enc -md sha256 -aes-256-cbc -in $1 -base64 -A  -k $GCLOUD_CREDENTIALS_KEY)
+
+VCS_TYPE=github
+CI_USERNAME=all-of-us
+CI_PROJECT=workbench
```
calbach: fix

ci/Dockerfile.circle_build
s-rubenstein: I believe the docker image above is what we've published to dockerhub? But that's just the base image? If I am wrong, let me know! I can fix.

s-rubenstein (commented):
Okay! I think I addressed all of your comments, except for one which I have a question for you as a follow up before addressing. @calbach when you have a chance, please take another look!! :D

s-rubenstein:
Also, do you know how to manually trigger these to run? Without creating side effects? (The deploy one, for example)

s-rubenstein:
Or with creating side effects and we just go for it, because messing with test and staging isn't really a big deal?

ci/Dockerfile.circle_build
calbach: Sorry, by above I was actually referring to the circle config file "- image: allofustest/workbench:buildimage-0.0.12". There's no reason to have this Dockerfile in multiple locations if they have the same exact image name. One will just overwrite the other when you publish the docker image.

calbach:
I'd just comment out the deploy thing for now while you are testing. It won't work until I fix it with my PR anyways

s-rubenstein:
@calbach should be commented out now.

ci/Dockerfile.circle_build
calbach: rm file?

s-rubenstein: Should be removed now. I was missing what the file was actually doing. (I thought it was loading in that docker image, not designed to publish it)

calbach (approved):

PR #8, opened 2018-11-29
Add tests back

s-rubenstein:


.circleci/config.yml
```diff
@@ -196,14 +196,14 @@ filter_releases: &filter_releases
...
+  build-test-deploy:
+    jobs:
+      # Always run basic test/lint/compilation (open PRs, master merge).
+      # Note: by default tags are not picked up.
+
+      # TODO: Add the public api testsapi tests
+      - public-ui-build-test
+      # Run deployment to "test" on master merges.
 #
 #      # TODO: Deploy public api to test
```
calbach: These functions *should* be working now. Do you want to try adding that in this PR?

s-rubenstein: Sure! Why not! Including the public-api-build-test?

s-rubenstein:
@calbach appears to be working! PTAL!

calbach (commented):
FYI: This commented out line is the culprit. I have no idea why it was commented out:
https://github.com/all-of-us/data-browser/blob/master/public-api/db-cdr/generate-cdr/init-new-cdr-db.sh#L33

Maybe @SrushtiGangireddy recalls (Peter made the change, but there were no relevant comments on the PR that I saw: https://github.com/all-of-us/workbench/pull/1203).

Anyways, looks like this is working with Circle env vars, though depending on that scares me a bit.

s-rubenstein:
@calbach PTAL when you get the chance! Appears to be fully working.

@SrushtiGangireddy if you are curious at all on the circle stuff.

SrushtiGangireddy:
Looks good. Thanks @blrubenstein . I understand some of it.

s-rubenstein:
@calbach, I tried uncommenting that, and it still fails for me. I think that fixed one, but didn't fix the MYSQL_ROOT_PASSWORD problem.

calbach (approved):
Sorry, I forgot to mention the other finding I had was that this command doesn't work locally (and doesn't work in workbench locally either).

PR #16, opened 2018-12-07
[risk=no] RW-1763 trimming cdr verison table

SrushtiGangireddy:
1. Removed unused columns from cdr version table.


common-api/src/main/java/org/pmiops/workbench/db/model/CdrVersion.java
```diff
@@ -20,12 +20,8 @@
   private boolean isDefault;
   private String name;
   private Short dataAccessLevel;
```
calbach: also don't need

public-api/db/changelog/db.changelog-master.xml
```diff
@@ -9,7 +9,8 @@
     <runningAs username="liquibase"/>
   </preConditions>
   <include file="changelog/db.changelog-1.xml"/>
+  <include file="changelog/db.changelog-2.xml"/>
```
calbach: The environments all already exist so it's probably too late to be changing the migration files around, aside from renaming files. I suppose we could delete the databases if we really wanted to... Did you have a plan for this?

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
```diff
@@ -13,7 +13,6 @@
 import org.pmiops.workbench.db.model.CommonStorageEnums;
```
calbach: Could you also remove this enum from the Swagger / enum converter utility?

calbach (commented):
Sorry - one last thing I noticed. Otherwise looks good

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
SrushtiGangireddy: Done. Thanks.

calbach: So I actually was asking whether you could remove these methods entirely:

- https://github.com/all-of-us/data-browser/blob/master/common-api/src/main/java/org/pmiops/workbench/db/model/CommonStorageEnums.java#L51
- (field + enum): https://github.com/all-of-us/data-browser/blob/master/common-api/src/main/resources/common_api.yaml#L65

SrushtiGangireddy: Was distracted working on something else. Pushed only the change from public-api.
Sorry.

common-api/src/main/resources/common_api.yaml
```diff
@@ -62,7 +62,6 @@ definitions:
     required:
       - cdrVersionId
       - name
-      - dataAccessLevel
```
calbach, line 65: Please remove the property and the enum and any other references in this file

common-api/src/main/resources/common_api.yaml
```diff
@@ -50,6 +45,7 @@ definitions:
       - defaultCdrVersionId
     properties:
       items:
+        type: "array"
```
calbach: revert?

calbach (approved):

PR #19, opened 2018-12-14
[RW-1809][risk=low] Add a separate requireSignIn config flag

calbach:
Previous PR did not fully drop the sign in requirement, just the IDV requirement.

PR #22, opened 2018-12-14
[risk=no] Fix login page logic when signin is not required

calbach:


SrushtiGangireddy (approved):
Looks good. Tested. Thanks.

s-rubenstein (commented):
You also need to change the API to allow for not having a key, and not needing registered access with workbench.

calbach:
@blrubenstein this was a follow-up omission from a previous PR: https://github.com/all-of-us/data-browser/pull/19

These two in combination should address the overall feature, let me know if you see any other outstanding issues.

PR #43, opened 2019-02-27
breadcrumb rebranding - 5

mhigbyflowers:
app/views/breadcrumb/component.html
app/views/breadcrumb/component.css
app/views/breadcrumb/component.ts


public-ui/src/app/views/breadcrumb/component.ts
```diff
@@ -0,0 +1,127 @@
...
+      if (label === 'Param: Cohort Name') {
+        label = child.snapshot.data['cohort'].name;
+      }
+      if (label === 'Param: Concept Set Name') {
+        label = child.snapshot.data['conceptSet'].name;
+      }
+      if (label === 'Param: Notebook Name') {
+        label = decodeURI(child.snapshot.params['nbName'])
+          .replace(/\.ipynb$/, '');
+      }
```
dolbeew: Looks like there's some workbench-specific code (the four if statements above and possibly elsewhere) that's not needed in data browser

public-ui/src/app/views/breadcrumb/component.html
```diff
@@ -0,0 +1,15 @@
+<div class="breadcrumb-container">
+  <div class="breadcrumb">
+    <a href="https://www.staging.researchallofus.org/">Home</a>
+  </div>
+  <div class="breadcrumb">
+    <a href="https://www.staging.researchallofus.org/data-research-tools/">Data & Reasearch Tools</a>
+  </div>
+  <div class="breadcrumb">
+    <a href="https://www.staging.researchallofus.org/data-research-tools/data-browser/">About The Databrowser</a>
+  </div>
```
dolbeew, line 10: Ideally these wouldn't be hardcoded, but looks like they're pointing to outside data browser so not sure what else would work. Maybe we could use a feature flag to point at the same environment you're currently using? Here's a [doc](https://docs.google.com/document/d/133JzAcrIf_ZKFqzL2jvVaBbxb69NMMSjs7iTuj8xW0A/edit#heading=h.pdgebek0f14m) CH created if you want to look into it.

dolbeew: Also, I haven't seen the mocks for this, but the breadcrumbs are styled differently than the research hub's:

Reasearch Hub
screen shot 2019-02-28 at 12 45 59 pm

Data Browser
screen shot 2019-02-28 at 12 46 13 pm

Not sure if those need to be consistent or not.

dolbeew (commented):
Left a few comments/questions. Also would recommend using the same naming convention as the other Data Browser components (breadcrumbs.component.ts) for consistency, and it prevents having a bunch of files with the same name.

public-ui/src/app/views/breadcrumb/component.html
dolbeew: Sorry, I think having an environment variable with the url for is fine instead of following the feature flag doc. Looks like each environment file already has urls for workbench and data browser for each environment ([workbench test url](https://github.com/all-of-us/data-browser/blob/0b8c8fed94f7678ebc2a19fc85708297ed07757e/public-ui/src/environments/environment.staging.ts#L4)), maybe can add variables for research hub sites also.

mhigbyflowers: image

PR #44, opened 2019-02-28
[DB-246] [risk=low] Fixing text repeat bug

SrushtiGangireddy:
1. Fixing the text repeat bug that places text over and over when searching for results.


mhigbyflowers:
Good solution using ngOnDistroy

mhigbyflowers (approved):
nice solution

PR #45, opened 2019-02-28
[risk=no] Display only valid age deciles in graph

SrushtiGangireddy:
1. cloning the pr from the workbench repo.


calbach (approved):
Ignore Codacy - I disabled this particular style rule (we already disabled it in Workbench).

PR #56, opened 2019-03-04
[risk=no] Handling cdr version name in init-new-cdr script #1871

SrushtiGangireddy:
Adding the version-flag in script to add necessary permission to public user based on the value in flag.

### DIFF
--- a/public-api/db-cdr/generate-cdr/init-new-cdr-db.sh
+++ b/public-api/db-cdr/generate-cdr/init-new-cdr-db.sh
@@ -5,9 +5,10 @@
 set -xeuo pipefail
 IFS=$'\n\t'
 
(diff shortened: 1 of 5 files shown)


public-api/db-cdr/generate-cdr/init-new-cdr-db.sh
```diff
@@ -32,8 +34,8 @@ fi
 # export for liquibase to use this
 #export CDR_DB_NAME
 
-# If CDR_DB_NAME matches ^public, we want to the public_db_user env var substituted in the create_db.sql
-if [[ CDR_DB_NAME =~ ^public ]]
+# If version flag contains public, we want to the public_db_user env var substituted in the create_db.sql
+if [[ $VERSION_FLAG =~ .*public.* ]]
```
calbach, line 38: Alternatively, could you remove this if check and the flags entirely? The CDR codepath should never be used in this repo.

PR #59, opened 2019-03-05
[risk=low] removing application type

SrushtiGangireddy:
1. Application type addition on post request causes the api to spin forever. so, removing it.


mhigbyflowers (approved):
👍

PR #62, opened 2019-03-05
Mhigbyflowers/rebranding

mhigbyflowers:


public-ui/src/environments/test-env-base.ts
```diff
@@ -6,5 +6,6 @@ export const testEnvironmentBase = {
   tcellapikey: 'AQEBBAEkx4iE2KxNyI7Wx08EwU1ycTM7E4FMSmaibbMUQxNU6uQvuAJt7fyABAtFYSYfgEE',
   publicApiUrl: 'https://api-dot-aou-db-test.appspot.com',
   workbenchUrl: 'https://all-of-us-workbench-test.appspot.com',
+  researchAllOfUsUrl: 'https://staging.researchallofus.org',
```
calbach, line 9: nit: can now remove this from `environment.ts` and `environment.local.ts` since they extend this base environment.

calbach (approved):
Looks good - please wait for Circle to go green before merging

public-ui/src/environments/test-env-base.ts
mhigbyflowers: thanks for the help

PR #64, opened 2019-03-05
[DB-251][risk=no] Update GA id for production

rushtong:
See https://precisionmedicineinitiative.atlassian.net/browse/DB-251

Update the GA ids to match the GA project we want this traffic to be reported in.

Traffic is currently broken down into three buckets:
* Local, Test, Staging
* Stable
* Production


SrushtiGangireddy (approved):
Looks good. Thanks.

calbach (approved):

PR #72, opened 2019-03-06
[risk=low] adding require signin only for prod, stable, staging

SrushtiGangireddy:
1. Not sure if this is how it can be addressed but added variables in environment file and check in sign-in-guard to only check sign in if require sign in is true.

### DIFF
--- a/public-ui/src/app/guards/sign-in-guard.service.ts
+++ b/public-ui/src/app/guards/sign-in-guard.service.ts
@@ -5,7 +5,7 @@ import {
   CanActivateChild,
   Router, RouterStateSnapshot
 } from '@angular/router';
-
+import {environment} from "../../environments/environment";
 import {Observable} from 'rxjs/Observable';
 
 import {SignInService} from 'app/services/sign-in.service';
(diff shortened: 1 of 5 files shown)


public-ui/src/app/guards/sign-in-guard.service.ts
```diff
@@ -18,6 +18,9 @@ export class SignInGuard implements CanActivate, CanActivateChild {
   constructor(private signInService: SignInService, private router: Router) {}
 
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
+    if (!environment.requireSignIn) {
```
calbach, line 21: Sorry - I misspoke when I said `environment.requireSignIn`. The value is currently defined on the server config.

Here is how it used to work - I don't know why it was removed: https://github.com/all-of-us/data-browser/blob/553b3bc55aaade44df9db4e4f7066f29b819cfcb/public-ui/src/app/guards/sign-in-guard.service.ts#L24

Could we just revert back to what we had previously? This is already out of sync with the server configs (requireSignIn is only enabled in prod right now). Unless there was some kind of performance reason for the change.

SrushtiGangireddy: got it. will do. thanks.

PR #55, opened 2019-03-04
Mhigbyflowers/rebrand touchups

mhigbyflowers:
Aesthetic changes + bug fixes

### Home Page
image

### EHR (conditions, procedures, measurements) 
image

### Surveys
image

### staging.researchallofus.org
Databrowser should transition close to as seemless as possible to staging researchallofus.org
image

image

PR #75, opened 2019-03-06
[risk=low] reverting back sign in change

SrushtiGangireddy:
1. Putting back in the change of sign in handling.

### DIFF
--- a/public-ui/src/app/guards/sign-in-guard.service.ts
+++ b/public-ui/src/app/guards/sign-in-guard.service.ts
@@ -8,6 +8,7 @@ import {
 
 import {Observable} from 'rxjs/Observable';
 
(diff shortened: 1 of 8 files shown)


SrushtiGangireddy:
Putting this back in. So, ch, when you say requireSignIn is only enabled in prod.
We would also want login for stable/ staging right ?

calbach (commented):
I think leave the configs for now due to appscan - though we probably do want stable to require sign in. I'll check this with the security team to verify.

I don't think this will fully fix it - this line was also changed and should be reverted : https://github.com/all-of-us/data-browser/pull/55/files#diff-e0fe15f640991f1289042b032d2f1dc8L1

PR #77, opened 2019-03-06
[risk=low] match on table name regex change

SrushtiGangireddy:
1. Changing regex to match the sub string in string.


mhigbyflowers (approved):
👍

PR #88, opened 2019-03-08
Mhf/route fix

mhigbyflowers:


public-ui/src/app/app-routing.module.ts
```diff
@@ -18,6 +18,16 @@ const routes: Routes = [
     component: LoginComponent,
     data: { title: 'Sign In' }
   },
+  {
```
mhigbyflowers: if user goes to /ehr or /survey they get redirected back to home-page

public-ui/src/app/app-routing.module.ts
```diff
@@ -18,6 +18,16 @@ const routes: Routes = [
...
+    path: 'ehr',
+    redirectTo: '',
```
mhigbyflowers, line 23: if user goes to /ehr or /survey they get redirected back to home-page

SrushtiGangireddy (approved):
Looks good.

PR #91, opened 2019-03-08
Mhf/ui tweaks3

mhigbyflowers:
general improvements.


mhigbyflowers:
had to close and reopen because some commits got stamped with the wrong user

PR #96, opened 2019-03-11
[risk=low] DB-349 race ethnicity charts display

SrushtiGangireddy:
1. Generated the data needed for race ethnicity charts and added them on the ui.


mhigbyflowers (approved):
👍

PR #98, opened 2019-03-11
[DB-304][risk=no] Performance Testing Infrastructure

rushtong:
Partially Addresses: https://precisionmedicineinitiative.atlassian.net/browse/DB-304

Things that this PR could do, but doesn't for time (will file follow-up tickets): 
* Circleci Job (auto-runs and manual runs)
* Generating tabular results suitable for db storage
* Storing results for historical comparisons
* Documentation: running an individual test
* Documentation: how to modify configurations for failure criteria
* Documentation: how to add a more complex user scenario
* Documentation: generalizing configuration settings


performance/README.md
```diff
@@ -0,0 +1,26 @@
+# Performance Testing with Gatling
+
+* [Gatling Documentation](https://gatling.io/docs/current/)
+* [Gradle Plugin Documentation](https://github.com/lkishalmi/gradle-gatling-plugin)
+
+## Running Tests
+
+From the `performance` directory, run the gatling tests (optionally accepts `-Denv=<ENV>` 
+which defaults to the test environment):
```
calbach: nit: could you give an example value here? e.g. "test" (or is the cloud project id?)

performance/src/gatling/simulations/org/pmiops/databrowser/Configuration.scala
```diff
@@ -0,0 +1,17 @@
...
+object Configuration {
+
+  val env: String = System.getProperty("env", "stable")
+  val defaultPause: Int = 3
+  val userAgentHeader: String = "Gatling Performance Testing"
+  val defaultUrl: String = {
+    env match {
+      case "prod" => "https://databrowser.researchallofus.org"
+      case "stable" => "https://www.databrowser.stable.fake-research-aou.org"
+      case "staging" => "https://www.databrowser.staging.fake-research-aou.org"
```
calbach: Per my comment further down, we should switch this to use the API endpoints instead, here are the correct values:

https://public.api.researchallofus.org
https://public.api.stable.fake-research-aou.org
https://api-dot-aou-db-staging.appspot.com
https://api-dot-aou-db-test.appspot.com

performance/src/gatling/simulations/org/pmiops/databrowser/Pages.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  object Home {
+    val home: ChainBuilder = exec(http(HomePage.name)
+      .get("/")
+      .check(status.is(session => 200)))
+      .pause(Configuration.defaultPause)
+  }
+
+  object Search {
+    val search: ChainBuilder = Home.home
```
calbach: I'm not quite following the dependency between Home and Search here. Is this basically saying "go to Home, then go to search?"

performance/src/gatling/simulations/org/pmiops/databrowser/Scenarios.scala
```diff
@@ -0,0 +1,23 @@
+package org.pmiops.databrowser
+
+import io.gatling.core.Predef.scenario
+import io.gatling.core.structure.ScenarioBuilder
+
+sealed trait UserScenario { val name: String }
```
calbach, line 6: Thanks for this code example, this makes things very concrete.

Now that I think I understand what Gatling does  - looks like it DOES NOT start a browser, e.g. have the ability to click a button on navigate a single page app. Assuming that's correct, lets not touch the UI at all here and go straight to the API. You may need to just observe what XHRs are being made as you walk through the user flow.

The UI service itself is purely static content, so there's not much point in load testing it. We're effectively load testing Google's static content service.

rushtong: You are correct - it is not in the same category as selenium, for example. Agreed on your suggested approach - that makes a lot of sense.

performance/src/gatling/simulations/org/pmiops/databrowser/Configuration.scala
rushtong: Will do. I'll also do some scenario refactoring so the focus is on APIs.

performance/src/gatling/simulations/org/pmiops/databrowser/Pages.scala
rushtong: It is exactly that. I agree this could be more clear - will do some more thinking along those lines as I do the scenario refactoring.

performance/README.md
rushtong: Added some more info here, hopefully it is clearer.

performance/src/gatling/simulations/org/pmiops/databrowser/Configuration.scala
rushtong: Done.

performance/src/gatling/simulations/org/pmiops/databrowser/Scenarios.scala
rushtong: Done.

calbach (approved):
Not sure whether this was ready for re-review, but the current code looks pretty nice to me, assuming it runs.

@SrushtiGangireddy FYI may also want to take a look through this

rushtong:
Thank you @calbach - I did mean to send out a `PTAL` comment at the end of yesterday but got sidetracked. Yes, this is the bare minimum I was hoping to get committed and then deal with additional testing requirements in subsequent PRs. 

@SrushtiGangireddy - Please do have a look and let me know if you have any questions.

SrushtiGangireddy:
Thanks @rushtong. I just glanced at this and i think i understand what's going on here.

PR #99, opened 2019-03-11
[risk=low] Fix number format exception

SrushtiGangireddy:
1. The value of string column in the stratum_4 throws number format exception when trying to be converted to float so adding try catch block to deal with it.

### DIFF
--- a/public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
+++ b/public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
@@ -830,16 +830,20 @@ public class DataBrowserController implements DataBrowserApiDelegate {
 
         for(AchillesResult ar: aa.getResults()){
             String analysisStratumName=ar.getAnalysisStratumName();
-            if(Long.valueOf(ar.getStratum3()) == MALE && maleBinRanges.contains(Float.parseFloat(ar.getStratum4()))){
-                maleBinRanges.remove(Float.parseFloat(ar.getStratum4()));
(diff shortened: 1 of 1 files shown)


public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -980,20 +984,24 @@ else if(Integer.parseInt(ard.getStratum3()) == '8') {
...
+                }else if(ar.getStratum2().equals("5") && binRanges5.contains(Float.parseFloat(ar.getStratum4()))){
+                    binRanges5.remove(Float.parseFloat(ar.getStratum4()));
+                }else if(ar.getStratum2().equals("6") && binRanges6.contains(Float.parseFloat(ar.getStratum4()))){
+                    binRanges6.remove(Float.parseFloat(ar.getStratum4()));
+                }else if(ar.getStratum2().equals("7") && binRanges7.contains(Float.parseFloat(ar.getStratum4()))){
+                    binRanges7.remove(Float.parseFloat(ar.getStratum4()));
+                }else if(ar.getStratum2().equals("8") && binRanges8.contains(Float.parseFloat(ar.getStratum4()))){
+                    binRanges8.remove(Float.parseFloat(ar.getStratum4()));
+                }
+            } catch (NumberFormatException nfe) {
```
calbach, line 1003: Is this really expected? Why? Please put a comment here or log a warning if it's actually not expected.

SrushtiGangireddy:
Closing out this pr, should not be needed.

PR #101, opened 2019-03-11
Mhf/delayed charts

mhigbyflowers:
When delaying the render of the chart for 1ms, the chart will fill its parent container.
This also fixes the issue where labels get cropped. 😃


SrushtiGangireddy (approved):
Looks good.

PR #111, opened 2019-03-13
[risk=no] changing age decile display names

SrushtiGangireddy:
1. Changing age decile display names on charts from "18-29","30-39",... -> <20,20,30,...


mhigbyflowers (approved):
👍

PR #114, opened 2019-03-13
[risk=low] updating survey question map file in generate-cdr/csv

SrushtiGangireddy:
1. Updating the survey question map csv file in generate-cdr/csv folder to match the one in db-cdr/csv.


mhigbyflowers (approved):
👍

PR #115, opened 2019-03-14
minor style updates

kathanner:
typo "sarif", margins update, removed inline chart styles


mhigbyflowers (approved):
Looks great. Thank You!

PR #118, opened 2019-03-14
[risk=no] Removing cdr version check

SrushtiGangireddy:
1. Removing cdr version check and unwanted csvs in generate-cdr/csv folder.


freemabd1 (approved):
LGTM

PR #131, opened 2019-03-18
Mhf/ie

mhigbyflowers:
adds support for Internet Explorer 11


dolbeew:
I _think_ you'll need to update `yarn.lock` manually with the new package so it matches `package.json` to get the test to pass. Also, I noticed your last commit desc was 'classlist.js not needed'. If that's the case, then remove it from `package.json` instead.

PR #137, opened 2019-03-18
Styles

kathanner:
updated bar charts style and removed side borders from nested tables like race/ethnicity from The Basics survey Question #2 https://aou-db-test.appspot.com/survey/The%20Basics


kathanner:
a few small style updates in bar chart and triple nested tables

mhigbyflowers (approved):
Looks good 👍 
thanks!

PR #140, opened 2019-03-19
nested survey table style

kathanner:
quick padding update for nested table to match https://docs.google.com/presentation/d/1-k64SkeHTKSra0CaL__BNj6U6sWCcEjYyPQm7mlutgs/edit#slide=id.g5298babc34_0_14


kathanner:
@mhigbyflowers -- updated margins and hover. Removed search icon from nav as a proposal to eliminate confusion.

mhigbyflowers (approved):
👍 Looks good. Let's think about other workarounds for the responsive table for the future. but we should go ahead and merge this so it can get on this release to production

PR #150, opened 2019-03-20
Revert "[DB-354][risk=low] Switch the prod UI to the prod API domain"

calbach:
Reverts all-of-us/data-browser#105

We're still waiting for the prod domains to update, so we need to revert this or prod will break.


SrushtiGangireddy (approved):
Looks good. Thanks

PR #156, opened 2019-03-20
breadcrumbs update for DB-403

kathanner:
removed text-transform on breadcrumbs, updated About the Data Browser crumb text


mhigbyflowers (approved):
👍

PR #160, opened 2019-03-21
[risk=no] deal with changing gender stratums

SrushtiGangireddy:
1. The api getGenderAnalysis was not returning results as the gender stratum nam ewas being set on wrong stratum. This pr correct that bug.

### DIFF
--- a/public-api/src/main/java/org/pmiops/workbench/cdr/dao/AchillesAnalysisDao.java
+++ b/public-api/src/main/java/org/pmiops/workbench/cdr/dao/AchillesAnalysisDao.java
@@ -18,7 +18,7 @@ public interface AchillesAnalysisDao extends CrudRepository<AchillesAnalysis, Lo
     )
     List<AchillesAnalysis> findConceptAnalysisResults(String concept_id,List<Long> analysisIds);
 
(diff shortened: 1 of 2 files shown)


public-api/src/main/java/org/pmiops/workbench/cdr/dao/AchillesAnalysisDao.java
```diff
@@ -18,7 +18,7 @@
     )
     List<AchillesAnalysis> findConceptAnalysisResults(String concept_id,List<Long> analysisIds);
 
-    @Query(value = "select a from AchillesAnalysis a left join FETCH a.results r where a.analysisId=?1")
+    @Query(value = "select distinct a from AchillesAnalysis a left join FETCH a.results as r where a.analysisId=?1")
```
calbach: nit: was this related?

calbach (approved):

public-api/src/main/java/org/pmiops/workbench/cdr/dao/AchillesAnalysisDao.java
SrushtiGangireddy: Changed it back. Merging in this pr.

PR #161, opened 2019-03-22
[DB-370][risk=low] Improve domain search perf

calbach:
This was the cleanest way I could find to force the query optimizer to do the right thing here. Basically the additional "or" predicates resulted in a full table scan, rather than using the full text index. In my direcdt SQL query experiments, this brought performance from ~8s -> ~10ms. Unfortunately the performance issue is not directly reproducible locally, since there are not as many synonyms loaded there.

In the future, we should move off FULLTEXT indexing since it seems to have a large number of performance issues / caveats.

Also adding @freemabd1  in case you have any additional MySQL insights here.

### DIFF
--- a/performance/src/gatling/simulations/org/pmiops/databrowser/Pages.scala
+++ b/performance/src/gatling/simulations/org/pmiops/databrowser/Pages.scala
@@ -28,7 +28,7 @@ object Pages {
     val domainTotals: HttpRequestBuilder = http("domain-totals").get("/v1/databrowser/domain-totals")
     def domainSearch (searchTerm: String): HttpRequestBuilder = {
       def encodedTerm: String = URLEncoder.encode(searchTerm, StandardCharsets.UTF_8.toString).toLowerCase
-      http("domain-totals").get("/v1/databrowser/domain-totals?searchWord=" + encodedTerm)
+      http("domain-search").get("/v1/databrowser/domain-search?searchWord=" + encodedTerm)
     }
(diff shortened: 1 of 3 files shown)


performance/src/gatling/simulations/org/pmiops/databrowser/Pages.scala
```diff
@@ -28,7 +28,7 @@ object Pages {
     val domainTotals: HttpRequestBuilder = http("domain-totals").get("/v1/databrowser/domain-totals")
     def domainSearch (searchTerm: String): HttpRequestBuilder = {
       def encodedTerm: String = URLEncoder.encode(searchTerm, StandardCharsets.UTF_8.toString).toLowerCase
-      http("domain-totals").get("/v1/databrowser/domain-totals?searchWord=" + encodedTerm)
+      http("domain-search").get("/v1/databrowser/domain-search?searchWord=" + encodedTerm)
```
calbach, line 31: @rushtong on further review I think this was just a typo as domain-totals doesn't take any parameters. Looks like the intent was to cover these domain keyword searches

freemabd1 (approved):
Looks reasonable to me. I completely agree with moving away from full text searches in mysql

SrushtiGangireddy (approved):
Looks good. Thanks. We changed or on the concept search and this was somehow not changed.

calbach:
Thanks. Per offline discussion it looks like the concept_code check was redundant (should be caught by the full text matching).

PR #163, opened 2019-03-22
[risk=moderate][DB-302]

mhigbyflowers:
## Set up for emergency
Created a boolean variable that if changed to false the site will only show /error page and won't allow any other navigation. 

### thumbnail of error page:
image

-- Also wasn't sure about the risk factor but I put 'moderate' because this is supposed to shut down the site if there is a data emergency

### DIFF
--- a/public-api/src/main/java/org/pmiops/workbench/publicapi/ConfigController.java
+++ b/public-api/src/main/java/org/pmiops/workbench/publicapi/ConfigController.java
@@ -27,6 +27,7 @@ public class ConfigController implements ConfigApiDelegate {
             .projectId(config.server.projectId)
             .enforceRegistered(config.auth.enforceRegistered)
             .requireSignIn(config.auth.requireSignIn)
+            .dataBrowserIsSafe(true) // change to false for emergency shutdown of frontend
             .publicApiKeyForErrorReports(config.server.publicApiKeyForErrorReports));
   }
 }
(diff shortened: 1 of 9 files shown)


public-api/src/main/java/org/pmiops/workbench/publicapi/ConfigController.java
```diff
@@ -27,6 +27,7 @@
             .projectId(config.server.projectId)
             .enforceRegistered(config.auth.enforceRegistered)
             .requireSignIn(config.auth.requireSignIn)
+            .dataBrowserIsSafe(true) // change to false for emergency shutdown of frontend
```
mhigbyflowers, line 30: This is the line that would be edited if there was an emergency

public-api/src/main/resources/public-api.yaml
```diff
@@ -319,6 +319,9 @@ definitions:
       requireSignIn:
         type: boolean
         description: Whether workbench sign in is required to use the data browser.
+      dataBrowserIsSafe:
```
mhigbyflowers, line 322: Variable declared in  public-api.yaml

public-ui/src/app/guards/is-safe-guard.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(private serverConfigService: ServerConfigService,
+    private router: Router) { }
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
+    return this.serverConfigService.getConfig().flatMap(config => {
+      // if true function and normal else show emergency page
+      if (config.dataBrowserIsSafe) {
+        return Observable.from([true]);
+      }
+        this.router.navigate(['/error']);
```
calbach: nit: unindent

calbach (approved):
Nice - thanks.

PR #164, opened 2019-03-25
[DB-421]  [risk=no] participant count fetch while searching in domains

SrushtiGangireddy:
1. Fetching total participant count when searching in domains.


mhigbyflowers (approved):
👍

PR #169, opened 2019-03-26
Db 290 rollup ui

mhigbyflowers:


SrushtiGangireddy (approved):
Looks good. Thanks. Wonder is it throws null exception on empty node / result items in recursive-tree.ts though.

PR #171, opened 2019-03-26
[risk=high] Emergency PR ( shutdown site with error-page )

mhigbyflowers:
# In case of an emergency, merge this pull request
dataBrowserIsSafe=false


calbach:
Nice tags - pre-approving in case it's needed.

calbach (approved):

PR #172, opened 2019-03-26
[Db-449] [risk=low] Updating count fetch logic

SrushtiGangireddy:
1. Update to count only the concepts that can be displayed.


mhigbyflowers (approved):
👍

PR #173, opened 2019-03-27
[risk=no][DB-290] source ui tweak

mhigbyflowers:
this add error message if no data found + checks if there is data to show


public-ui/src/app/views/ehr-view/ehr-view.component.css
```diff
@@ -140,6 +140,41 @@ h5.secondary-display {
     font-size:1.3em;
 }
 
+.loading-dots {
```
SrushtiGangireddy: if this is the same as above. could this be moved into a common place ?

public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -184,11 +185,12 @@ export class EhrViewComponent implements OnInit, OnDestroy {
     this.resetSelectedGraphs();
     this.graphToShow = g;
     if (this.graphToShow === GraphType.Sources) {
+      this.treeLoading = true;
       this.subscriptions.push(this.api.getCriteriaRolledCounts(r.conceptId)
-        .subscribe({
+      .subscribe({
           next: result => {
             this.treeData = [result.parent];
-            this.treeData['children'] = result.children.items;
```
SrushtiGangireddy, line 191: where is this being set now ?

mhigbyflowers: It was redundant. It didn't need it to get the next level children.

public-ui/src/app/views/ehr-view/ehr-view.component.css
mhigbyflowers: Yep. I will add it to the loading component.

SrushtiGangireddy (approved):
Looks good. thanks.

PR #181, opened 2019-03-28
[DB-450] [risk=no] updating the % of participants header

SrushtiGangireddy:
1. Updating the % of total participants to % of xx,xxx whatever is the domain count.
2. Concept charts was throwing error on first concept of measurement and procedure saying cannot find results of null.
3. Removing scroll into view as this was causing issues when multiple graphs are opened.


mhigbyflowers (approved):
👍

PR #188, opened 2019-03-29
[risk=no][DB-396] show more bug

mhigbyflowers:
fixed 1. Diabetes mellitus” is where See More doesn’t do anything


SrushtiGangireddy (approved):
Looks good. 
Also, 
https://github.com/all-of-us/data-browser/blob/master/public-ui/src/app/views/ehr-view/ehr-view.component.html#L212 this line needs change may be to align with the rest.
Thanks.

PR #194, opened 2019-04-01
[DB-370][risk=no] Load test tweaks

calbach:
- Add a surveys scenario
- Disable caching
- Switch to constant user traffic
- Use the traffic variables across all scenarios


calbach:
hmm, forgot about this - please take a look when you can @SrushtiGangireddy

SrushtiGangireddy (approved):
Sorry for delay. Thanks looks good to me.

PR #195, opened 2019-04-01
[DB-457] [risk=no] adding tooltip to questions that have multiple answers

SrushtiGangireddy:
1. Adding tooltip to survey questions that have multiple answers.
2. Dealing with null exception.


mhigbyflowers (approved):
👍 thanks

PR #204, opened 2019-04-02
[DB-351] [risk=no] select the searched pm

SrushtiGangireddy:
1. Select the searched physical measurement.
2. Display the count of matched pms.


mhigbyflowers (approved):
looks good 👍

PR #207, opened 2019-04-04
[DB-475] [risk=no] adding concept code check when fetching domain counts

SrushtiGangireddy:
1. Adding the concept code check (check concept code and fetch it even if it is not standard concept).

### DIFF
--- a/common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
+++ b/common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
@@ -36,11 +36,11 @@ public interface DomainInfoDao extends CrudRepository<DomainInfo, Long> {
       // move these searches to a more suitable technology, e.g. Elasticsearch.
       "    UNION DISTINCT " +
       "    (select domain_id, concept_id from concept " +
-      "     where concept_id in (?2) and can_select=1)) c1 " +
+      "     where concept_id in (?3) or concept_code = ?2 and can_select=1)) c1 " +
       "  group by c1.domain_id) c " +
       "ON d.domain_id = c.domain_id " +
       "order by d.domain_id")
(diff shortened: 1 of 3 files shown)


public-api/src/main/resources/public-api.yaml
```diff
@@ -61,15 +61,15 @@ paths:
     get:
       tags:
         - dataBrowser
-
       description: Gets the cdr versions
       operationId: "getCdrVersionUsed"
+      security: []
```
SrushtiGangireddy, line 66: image
This is to deal with this error.
Not sure if it helps though.
I do not see this error on my local server.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
```diff
@@ -36,11 +36,11 @@
       // move these searches to a more suitable technology, e.g. Elasticsearch.
       "    UNION DISTINCT " +
       "    (select domain_id, concept_id from concept " +
-      "     where concept_id in (?2) and can_select=1)) c1 " +
+      "     where concept_id in (?3) or concept_code = ?2 and can_select=1)) c1 " +
```
SrushtiGangireddy, line 39: sorry @calbach. I totally forgot about this when you asked me the other day.
So, the idea was to show the concept count when searched with concept code irrespective of they being standard. 
The other search check needs the standard concept check.
So added it back in here.

calbach (approved):

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
calbach: Ah, because of the `and standard_concept IN ('S', 'C')` above

PR #211, opened 2019-04-05
DB-419/risk=low/update nav, header, footer, breadcrumbs

kathanner:


kathanner:
@mhigbyflowers -- can you review this for the updated navigation, header style, footer and breadcrumbs? One issue is just a css "!important" and the other is with the .ts file we worked through together about whitespace. I'm not sure what that's referring to.

mhigbyflowers (approved):
👍  looks good thanks

PR #213, opened 2019-04-05
[risk=no] Upgrade docker image, includes new GAE version, revert web.xml

calbach:
My original PR was "correct", but our circle image was using an old version of the GAE SDK, which is why the property name failed originally. I rebuilt the image and it now matches we have locally.


SrushtiGangireddy (approved):
Looks good. thanks.

PR #214, opened 2019-04-05
[DB-343] [risk=low] Search keyword tracking in google analytics

SrushtiGangireddy:
1. Triggering the post call to post event data to analytics.
2. Added @Async on the method call for asynchronous execution.

### DIFF
--- a/public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsService.java
+++ b/public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsService.java
@@ -6,6 +6,6 @@ import java.io.IOException;
 public interface GoogleAnalyticsService {
 
     int trackEventToGoogleAnalytics(@Nullable String cid, String category, String action,
-                                    String label, String value) throws IOException;
+                                    String label) throws IOException;
 
 }
--- a/public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsServiceImpl.java
+++ b/public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsServiceImpl.java
@@ -59,7 +59,7 @@ public class GoogleAnalyticsServiceImpl implements GoogleAnalyticsService {
      */
     @Override
     public int trackEventToGoogleAnalytics(@Nullable String cid, String category, String action,
-                                           String label, String value) throws IOException {
+                                           String label) throws IOException {
         String defaultCid = Optional.ofNullable(cid).orElse("555");
         Preconditions.checkNotNull(category);
         Preconditions.checkNotNull(action);
@@ -71,7 +71,6 @@ public class GoogleAnalyticsServiceImpl implements GoogleAnalyticsService {
         map.put("ec", encode(category));
         map.put("ea", encode(action));
         map.put("el", encode(label));
-        map.put("ev", encode(value));
 
         HTTPRequest request = new HTTPRequest(getGoogleAnalyticsEndpoint(), HTTPMethod.POST);
         request.addHeader(this.header);
--- a/public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsServiceImpl.java
+++ b/public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsServiceImpl.java
@@ -5,7 +5,7 @@ import com.google.appengine.api.urlfetch.HTTPMethod;
 import com.google.appengine.api.urlfetch.HTTPRequest;
 import com.google.appengine.api.urlfetch.HTTPResponse;
 import com.google.appengine.api.urlfetch.URLFetchService;
-import com.google.appengine.repackaged.com.google.common.base.Preconditions;
+import com.google.common.base.Preconditions;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
--- a/public-api/src/main/java/org/pmiops/workbench/publicapi/AsyncConfiguration.java
+++ b/public-api/src/main/java/org/pmiops/workbench/publicapi/AsyncConfiguration.java
@@ -21,4 +21,5 @@ public class AsyncConfiguration
         executor.initialize();
         return executor;
     }
+
 }
--- a/public-api/src/main/java/org/pmiops/workbench/publicapi/AsyncConfiguration.java
+++ b/public-api/src/main/java/org/pmiops/workbench/publicapi/AsyncConfiguration.java
@@ -9,8 +9,7 @@ import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
 @Configuration
 @EnableAsync
-public class AsyncConfiguration
-{
+public class AsyncConfiguration {
     @Bean(name = "asyncExecutor")
     public Executor asyncExecutor() {
         ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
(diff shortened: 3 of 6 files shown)


public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsService.java
```diff
@@ -6,6 +6,6 @@
 public interface GoogleAnalyticsService {
 
     int trackEventToGoogleAnalytics(@Nullable String cid, String category, String action,
-                                    String label, String value) throws IOException;
+                                    String label) throws IOException;
```
SrushtiGangireddy, line 9: According to https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#ev
the event value seems to take only integer value. So passing in the search term as label.

public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsServiceImpl.java
```diff
@@ -5,7 +5,7 @@
 import com.google.appengine.api.urlfetch.HTTPRequest;
 import com.google.appengine.api.urlfetch.HTTPResponse;
 import com.google.appengine.api.urlfetch.URLFetchService;
-import com.google.appengine.repackaged.com.google.common.base.Preconditions;
```
SrushtiGangireddy, line 8: This was throwing an error somehow, so changed it.

public-api/src/main/java/org/pmiops/workbench/publicapi/AsyncConfiguration.java
```diff
@@ -0,0 +1,24 @@
+package org.pmiops.workbench.publicapi;
```
SrushtiGangireddy, line 1: This is executor method that is used by methods annotated by @Async

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -87,6 +92,8 @@
     private Provider<CdrVersion> defaultCdrVersionProvider;
     @Autowired
     private ConceptService conceptService;
+    @Autowired
+    private GoogleAnalyticsServiceImpl googleAnalyticsService;
```
rushtong: Do you want the interface here, or does that not autowire correctly to the impl?

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -1256,4 +1266,15 @@ else if(Integer.parseInt(ard.getStratum3()) == '8') {
             aa.addResult(achillesResult);
         }
     }
+
+    @Async("asyncExecutor")
+    public Future<String> searchTrackEvent(String category, String event, String label) {
+        int response = -1;
+        try{
```
rushtong: nit: spacing
```suggestion
        try {
```

public-api/src/main/java/org/pmiops/workbench/google/GoogleAnalyticsService.java
rushtong: Nice catch!

public-api/src/main/java/org/pmiops/workbench/publicapi/AsyncConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+    public Executor asyncExecutor() {
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        executor.setCorePoolSize(3);
+        executor.setMaxPoolSize(3);
+        executor.setQueueCapacity(100);
+        executor.setThreadNamePrefix("AsynchThread-");
+        executor.initialize();
+        return executor;
+    }
+}
```
rushtong, line 24: nit: newline after brace

rushtong (approved):
👍 This looks reasonable to me. Minor comments inline.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: It is the implementation being autowired. Bad named it with the interface name. Sorry. Corrected.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: Addressed. thanks.

public-api/src/main/java/org/pmiops/workbench/publicapi/AsyncConfiguration.java
SrushtiGangireddy: Addressed. thanks.

public-api/src/main/java/org/pmiops/workbench/publicapi/AsyncConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.EnableAsync;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+@Configuration
+@EnableAsync
+public class AsyncConfiguration
+{
```
calbach: nit: bracket to prev line - I don't think we do this style anywhere else

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -1256,4 +1266,15 @@ else if(Integer.parseInt(ard.getStratum3()) == '8') {
...
         }
     }
+
+    @Async("asyncExecutor")
+    public Future<String> searchTrackEvent(String category, String event, String label) {
+        int response = -1;
+        try {
+            response = googleAnalyticsService.trackEventToGoogleAnalytics(null, category, event, label);
+        } catch (IOException ioException) {
+            response = -1;
```
calbach: This should be redundant with line 1272. `trackEventToGoogleAnalytics` cannot both return a value and throw an exception

I would also suggest logging a warning here, or this information will be lost

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -493,6 +500,7 @@ public SurveyQuestionResult apply(AchillesResult o) {
             toMatchConceptIds.addAll(drugMatchedConcepts.stream().map(Concept::getConceptId).collect(Collectors.toList()));
         }
 
+        searchTrackEvent("Search", "DomainSearch", query);
```
calbach: Does this really happen asynchronously? If so Spring is doing some insane black magic. From scanning through the docs I wasn't quite sure if this would work, or if it would need to be a call made to another `@Service` (or spring-instantiated object). Could you confirm from the logs?

(also can print `Thread.currentThread().getName()` here and in `searchTrackEvent` to confirm)

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -1256,4 +1266,15 @@ else if(Integer.parseInt(ard.getStratum3()) == '8') {
             aa.addResult(achillesResult);
         }
     }
+
+    @Async("asyncExecutor")
+    public Future<String> searchTrackEvent(String category, String event, String label) {
```
calbach: opt: This could also just be void, as no one is using the results

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: It gives different thread names.

| qtp1262773598-18      
 | qtp1262773598-16      
| qtp1262773598-22

calbach: Neat - thanks for checking.

calbach (approved):

SrushtiGangireddy:
Thanks merging in.

PR #215, opened 2019-04-05
[DB-494] [risk=no] Make whole survey answer row clickable for sub level answers

SrushtiGangireddy:
1. Few graphs in third level were not showing up, dealing with that.


mhigbyflowers (approved):
👍 thanks

PR #219, opened 2019-04-08
[risk=no][DB-458]urls match titles

mhigbyflowers:
the cards:
Data Browser > Conditions
Data Browser > Drug Exposures
Data Browser > Measurements
Data Browser > Procedures

And the URLS also to match the cards.
.com/ehr/conditions
.com/ehr/drug-exposures
.com/ehr/measurements
.com/ehr/procedures


public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -84,6 +85,14 @@ export class EhrViewComponent implements OnInit, OnDestroy {
     }
   }
 
+  public getDomainFromRoute(routeParam: string) {
+    this.dbc.routeToDomainMap.forEach(item => {
+      if (routeParam === item.route) {
+        console.log(routeParam, item.route);
```
SrushtiGangireddy: this can be removed if debugging is done may be.

public-ui/src/app/views/quick-search/quick-search.component.ts
```diff
@@ -202,16 +213,24 @@ export class QuickSearchComponent implements OnInit, OnDestroy {
   public viewEhrDomain(r) {
     localStorage.setItem('ehrDomain', JSON.stringify(r));
     localStorage.setItem('searchText', this.prevSearchText);
-    this.router.navigateByUrl('ehr/' + r.domain.toLowerCase().replace(' ', '-' ));
+    for (const item in this.dbc.routeToDomainMap) {
```
SrushtiGangireddy: Instead of having for each multiple times, could we have two hashmaps routeToDomain and domainToRoute and then pick the domain/ route by key. It is just four elements so does not make any difference probably.

SrushtiGangireddy (approved):
Looks good with just a comment, thanks.

public-ui/src/app/views/quick-search/quick-search.component.ts
mhigbyflowers: Great suggestion. thanks

PR #220, opened 2019-04-08
[DB-343] [risk=no] send page path along with the event params

SrushtiGangireddy:
1. Send page path to show up in the page parameter along with the search term in event label.

### DIFF
--- a/public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
+++ b/public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
@@ -425,7 +425,7 @@ public class DataBrowserControllerTest {
     public void testAllConceptSearchEmptyQuery() throws Exception{
         ResponseEntity<ConceptListResponse> response = dataBrowserController.searchConcepts(new SearchConceptsRequest()
                 .domain(Domain.CONDITION)
-                .minCount(0),"test/ehr/condition");
+                .minCount(0), "test/ehr/condition");
       assertThat(response.getBody().getItems()).containsExactly(CLIENT_CONCEPT_1, CLIENT_CONCEPT_5,
           CLIENT_CONCEPT_6, CLIENT_CONCEPT_7);
     }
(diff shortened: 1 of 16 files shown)


public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
```diff
@@ -435,36 +435,36 @@ public void testConceptSearchMaxResults() throws Exception{
...
-                .domain(Domain.CONDITION));
+                .domain(Domain.CONDITION),"test/ehr/condition");
         assertThat(response.getBody().getItems()).containsExactly(CLIENT_CONCEPT_1, CLIENT_CONCEPT_5, CLIENT_CONCEPT_6);
     }
 
     @Test
     public void testConceptSearchStandardConcept() throws Exception{
         ResponseEntity<ConceptListResponse> response = dataBrowserController.searchConcepts(new SearchConceptsRequest().query("002")
-                .standardConceptFilter(StandardConceptFilter.STANDARD_CONCEPTS));
+                .standardConceptFilter(StandardConceptFilter.STANDARD_CONCEPTS),"test/ehr/condition");
```
calbach: nit: space between parameters (we should have a linter for this)

public-ui/src/app/components/db-no-results/db-no-results.component.ts
```diff
@@ -84,7 +85,8 @@ export class DbNoResultsComponent implements OnChanges, OnDestroy {
   }
 
   public searchDomains(query: string) {
-    this.subscriptions.push(this.api.getDomainSearchResults(query).subscribe(results => {
+    this.subscriptions.push(this.api.getDomainSearchResults(query,
+      environment.publicUiUrl + this.router.url).subscribe(results => {
```
calbach: I don't think including the full URL is necessary here, seems like just the url should do

public-ui/src/environments/environment.prod.ts
```diff
@@ -2,6 +2,7 @@ export const environment = {
   displayTag: '',
   publicApiUrl: 'https://public.api.researchallofus.org',
   workbenchUrl: 'https://workbench.researchallofus.org',
+  publicUiUrl: 'http://aou-db-prod.appspot.com/',
```
calbach: I don't think this is necessary per above

public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
SrushtiGangireddy: Added. Yes linter in space would help. Not sure how to add it though.

public-ui/src/app/components/db-no-results/db-no-results.component.ts
SrushtiGangireddy: image
Was adding it as helen asked for this. I agree with not needing this though since there is a way to view events seperated by url by selecting the project on the top.

calbach: I see the concern, the event category should be enough here, but if you really want the page could you just use `window.location.href` instead of the environment flag?

SrushtiGangireddy: Changed it to use window.location.href instead. thanks.

calbach (approved):

PR #222, opened 2019-04-08
[DB-484] [risk=no] Fetch only unique concepts

SrushtiGangireddy:
1. Concept search + drug search is fetching duplicates. So remove the ones fetched with concept search in drug search.


mhigbyflowers (approved):
👍 look good

PR #224, opened 2019-04-09
[DB-343] [risk=no] survey page search event 

SrushtiGangireddy:
The survey pages fetch all the questions and answers before hand and then filter locally, so there is no way the search term change on client side to track. So hitting api to trigger the event with the empty response.


calbach (commented):
I'm not following - what do you mean by not being able to trigger the event from the client side? I'd rather do that than overload an unrelated method for tracking purposes.

This should work fine:
`gtag('send', 'event', [eventCategory], [eventAction], [eventLabel], [eventValue], [fieldsObject]);`

https://developers.google.com/analytics/devguides/collection/analyticsjs/events

PR #225, opened 2019-04-09
[DB-496] [risk=no] Display the male and female participant counts on value graphs

SrushtiGangireddy:
1. Display the male and female participant counts on value graphs.
2. Solidify the chart icon when graphs are expanded in surveys.


mhigbyflowers (approved):
👍 looks good

PR #228, opened 2019-04-09
[DB-343] [risk=no] survey search track event:

SrushtiGangireddy:
1. Triggering the search term event each time the search term changes on the client side.

### DIFF
--- a/public-ui/src/app/views/survey-view/survey-view.component.ts
+++ b/public-ui/src/app/views/survey-view/survey-view.component.ts
@@ -268,11 +268,9 @@ export class SurveyViewComponent implements OnInit, OnDestroy {
     if (this.searchText.value.length > 0) {
       this.questions = this.questions.filter(this.searchQuestion, this);
       gtag('set', 'page', window.location.href);
-      gtag('send', {
-        hitType: 'event',
-        eventCategory: 'DataBrowserSearch',
-        eventAction: 'SurveySearch',
-        eventLabel: this.searchText.value,
+      gtag('event', 'SurveySearch', {
+        'event_category': 'DataBrowserSearch',
+        'event_label': this.searchText.value
       });
     }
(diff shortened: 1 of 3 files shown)


public-ui/src/app/views/survey-view/survey-view.component.ts
```diff
@@ -262,6 +267,11 @@ export class SurveyViewComponent implements OnInit, OnDestroy {
     }
     if (this.searchText.value.length > 0) {
       this.questions = this.questions.filter(this.searchQuestion, this);
+      gtag('set', 'page', window.location.href);
```
calbach: Could you double check whether you can remove this? I would guess this would be getting propagated automatically.

calbach (approved):
Thanks!

public-ui/src/app/views/survey-view/survey-view.component.ts
```diff
@@ -46,6 +50,7 @@ export class SurveyViewComponent implements OnInit, OnDestroy {
   constructor(private route: ActivatedRoute, private api: DataBrowserService,
     private tooltipText: TooltipService,
     public dbc: DbConfigService) {
+    gtag('config', environment.gaId, 'auto');
```
calbach: Could you check whether you can remove this? This seems like a global setting being applied in this very specific page component.

SrushtiGangireddy: Yes, it works when i remove this. Thanks.

public-ui/src/app/views/survey-view/survey-view.component.ts
SrushtiGangireddy: Page is being auto tracked irrespective of this. thanks.

SrushtiGangireddy:
Thanks, merging in.

PR #231, opened 2019-04-09
[DB-467] [risk=no] Updating auto insert part in survey questions

SrushtiGangireddy:
1. Updating auto insert part in survey questions


mhigbyflowers (approved):
👍 thanks

PR #234, opened 2019-04-10
[DB-454] [risk=no] Removing race/ethnicity tab on the graphs

SrushtiGangireddy:
1. Removing race/ ethnicity charts.
2. Stopping propagation on the header events as link were not clickable under the main links.


mhigbyflowers (approved):
👍 looks good

PR #236, opened 2019-04-10
[DB-523] [risk=no] Removing branching logic for race/ ethnicity question in basics.

SrushtiGangireddy:
1. Removing branching logic for race/ ethnicity question in basics.


mhigbyflowers (approved):
👍 thanks

PR #238, opened 2019-04-10
[DB-361] [risk=no] Update the search term in the url 

SrushtiGangireddy:
1. update the search term in the url to share urls

### DIFF
--- a/public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
+++ b/public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
@@ -578,8 +578,12 @@ public class DataBrowserController implements DataBrowserApiDelegate {
         List<Concept> conceptList = new ArrayList(concepts.getContent());
 
         if(searchConceptsRequest.getDomain() != null && searchConceptsRequest.getDomain().equals(Domain.DRUG) && !searchConceptsRequest.getQuery().isEmpty()) {
-            List<Concept> drugMatchedConcepts = conceptDao.findDrugIngredientsByBrandNotInConceptIds(searchConceptsRequest.getQuery(), conceptList.stream().map(Concept::getConceptId).collect(Collectors.toList()));
-
+            List<Concept> drugMatchedConcepts = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -578,8 +578,12 @@ public SurveyQuestionResult apply(AchillesResult o) {
         List<Concept> conceptList = new ArrayList(concepts.getContent());
 
         if(searchConceptsRequest.getDomain() != null && searchConceptsRequest.getDomain().equals(Domain.DRUG) && !searchConceptsRequest.getQuery().isEmpty()) {
-            List<Concept> drugMatchedConcepts = conceptDao.findDrugIngredientsByBrandNotInConceptIds(searchConceptsRequest.getQuery(), conceptList.stream().map(Concept::getConceptId).collect(Collectors.toList()));
-
+            List<Concept> drugMatchedConcepts = new ArrayList<>();
+            if (conceptList.size() > 0) {
+                drugMatchedConcepts = conceptDao.findDrugIngredientsByBrandNotInConceptIds(searchConceptsRequest.getQuery(), conceptList.stream().map(Concept::getConceptId).collect(Collectors.toList()));
+            } else {
+                drugMatchedConcepts = conceptDao.findDrugIngredientsByBrand(searchConceptsRequest.getQuery());
```
SrushtiGangireddy, line 585: in on empty list in sql query was throwing error. So in case there are no concepts using the one without in.

mhigbyflowers (approved):
Cool! I didn't know about queryParams.
I would suggest that we use '?search=' instead of '?searchString' unless there is a reason that I am unaware of, it seems cleaner to me.
I wonder if we can expand and use something like '?icd9='? Is the search set up to use codes?

SrushtiGangireddy:
Changed search string to search. Not sure if it can be changed to icd9= since the search has various attrs (concept name/ concept code/ concept id).

mhigbyflowers (approved):
👍reason for not having icd9= makes sense to me

PR #239, opened 2019-04-10
[DB-525][risk=low] Only require signin for prod databrowser access

calbach:
Preparing this PR in case we move ahead with this option - needs security approval.

To apply this:
1. Merge thhis PR
2. Locally, pull master
3. `./project.rb update-cloud-config --project aou-db-prod`

PR #240, opened 2019-04-10
[DB-525][risk=low] Fix enforceRegistered check

calbach:
Check dataAccessLevel rather than the deprecated idVerificationStatus field.

See also #239 , which makes this a non-issue (the code should still be updated here to match reality).


calbach:
Verified this fix locally.

SrushtiGangireddy (approved):
Looks good. thanks.

PR #241, opened 2019-04-11
[risk=no][DB-524]scroll bug

mhigbyflowers:
removing scrollto elements: fixed bug


SrushtiGangireddy (approved):
Looks good, thanks.
Left it there forgot to remove.

PR #242, opened 2019-04-11
[DB-523] [risk=no] Removing branching logic of rest of answers as it is 

SrushtiGangireddy:
1. Changed the data to not show the branching logic of AI/ AN and pacific islander.
2. The decision of nih was to show the rest of answer graphs without race/ ethnicity.


mhigbyflowers (approved):
👍

PR #243, opened 2019-04-11
[risk=no] removing rows in survey question map to generate prod data

SrushtiGangireddy:
1.  Removed American indian/ alaskan native and pacific islander sub question rows so that the data would not be pulled up.


mhigbyflowers (approved):
👍 thanks

PR #244, opened 2019-04-11
[risk=no][DB-284]Scroll only when window is small + bold highlight

mhigbyflowers:


SrushtiGangireddy (approved):
Looks good. nice catch. thanks.

SrushtiGangireddy (approved):
Looks good. nice catch. thanks.

SrushtiGangireddy (commented):
Looks good. nice catch. thanks.

PR #246, opened 2019-04-11
[DB-361] [risk=no] removing the update search url changes and adding new prod config

SrushtiGangireddy:
1. Added new prod cdr version in config.
2. Removing the search param in the query param keeping sql injection attack in mind. Will revisit later.


mhigbyflowers (approved):
👍

PR #251, opened 2019-04-12
[DB-528] [risk=low] Fix the query that adds non-present breakdown counts 

SrushtiGangireddy:
1. Fix the query that adds non-present breakdown charts.


mhigbyflowers (approved):
👍 looks good.
idk if stopping prop fixes the problem. Im looking in to it.

SrushtiGangireddy:
Okay pushing this on test. We can think of something else if this still does not work out thanks.

PR #252, opened 2019-04-12
[risk=no][DB-502]source-tree-bugs

mhigbyflowers:
tree works much better. only one sibling can be opened at a time but it works better
also got rid of the spacing bug.

### DIFF
--- a/public-ui/src/app/components/recursive-tree/recursive-tree.component.ts
+++ b/public-ui/src/app/components/recursive-tree/recursive-tree.component.ts
@@ -36,11 +36,13 @@ export class RecursiveTreeComponent implements OnChanges, OnDestroy {
     }
   }
   openUp(index, node, nodes) {
+    console.log(index, node, nodes);
+    
     for (let i = 0; i < nodes.length; i++) {
       const item = nodes[i];
       if (item === node && i === index) {
         this.childIndex = i;
-        this.subOpened = !this.subOpened;
+        this.subOpened = true;
       }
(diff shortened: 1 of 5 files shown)


public-ui/src/app/components/recursive-tree/recursive-tree.component.ts
```diff
@@ -40,7 +40,7 @@ export class RecursiveTreeComponent implements OnChanges, OnDestroy {
       const item = nodes[i];
       if (item === node && i === index) {
         this.childIndex = i;
-        this.subOpened = !this.subOpened;
+        this.subOpened = true;
```
SrushtiGangireddy, line 43: when would this change to false on closing ?

mhigbyflowers: the component starts as false.
and when another node is clicked it sends the 'opened' var as false, which is used to render/unrender the children'

SrushtiGangireddy (approved):

public-ui/src/app/components/recursive-tree/recursive-tree.component.ts
mhigbyflowers: kinda hard to explain because of recursion

SrushtiGangireddy: ohh yeah gotcha. Was circling over there. thanks.

PR #255, opened 2019-04-12
[DB-240] [risk=no] making the whole box clickable

SrushtiGangireddy:
1. Making the whole of domain info display box clickable instead of just the link view results in the bottom.


mhigbyflowers (approved):
👍

PR #262, opened 2019-04-15
[DB-536] [risk=no] tooltip for q2 rolled into other

SrushtiGangireddy:
1. Helptext for the category "other" in q2 to describe the fact that it has counts rolled up from american indian and such categories.


mhigbyflowers (approved):
👍

PR #264, opened 2019-04-16
[DB-538] [risk=no] matching underlined name on survey charts

SrushtiGangireddy:
1. Matching the underlined name on survey charts when multiple charts are opened.


mhigbyflowers (approved):
👍 thanks

PR #268, opened 2019-04-17
[DB-543] [risk=no] change age decile label name

SrushtiGangireddy:
1. Change the age decile label to be consistent with the data.


mhigbyflowers (approved):
👍 thanks

PR #269, opened 2019-04-17
DB#[risk=low]match WP site

kathanner:


kathanner:
@mhigbyflowers -- this is the update to match (as best as possible) the WP site header and footer updates.

mhigbyflowers (approved):
Looks good. I added a (click)="$event.stopPropagation();" to a few anchor tags to keep the submenus from collapsing.

PR #271, opened 2019-04-18
[DB-557] [risk=no] Change the biological sex label in charts

SrushtiGangireddy:
1. Change the biological sex chart label.
2. Change in help text.


mhigbyflowers (approved):
👍 looks good

PR #275, opened 2019-04-19
[DB-543] [risk=no] Changing the count in missing age decile from 0 - 20

SrushtiGangireddy:
1. We are adding the missing age deciles with empty counts to smooth the look of histogram.
2. Changing the count from 0 - 20.


mhigbyflowers (approved):
👍

PR #278, opened 2019-04-19
[DD-550] [risk=low] Remove the bad values form overall health q10

SrushtiGangireddy:
1. Remove bad integer values from overall health q10.


mhigbyflowers (approved):
👍 looks good

PR #279, opened 2019-04-19
[DB-550] [risk=no] Remove bad values in q10 in overall health

SrushtiGangireddy:
q. Remove the values that are out of bounds in q10.


mhigbyflowers (approved):
👍

PR #282, opened 2019-04-22
[DB-555] [risk=no] sorting the answers of subquestions in the order of their participant…

SrushtiGangireddy:
1. Sorting the answers of sub-questions in the order of their participant counts.


mhigbyflowers (approved):
👍 thanks

PR #286, opened 2019-04-23
[DB-560] [risk=no] count the drug concepts with count > 0

SrushtiGangireddy:
1. only count drug concepts with count > 0.


mhigbyflowers (approved):
👍 looks good

PR #294, opened 2019-04-26
[DB-573] [risk=no] adding less than symbol for label on charts

SrushtiGangireddy:
1. Adding actual unicode less than or equal to symbol on charts.


SrushtiGangireddy:
@mhigbyflowers  Just adding usehtml in there and letting style be as is fetches an extra box in there. 
You can change it if you have better idea to deal with that.

mhigbyflowers (approved):
👍

PR #297, opened 2019-04-29
[DB-227] [risk=no] Preserve search term in session

SrushtiGangireddy:
1. Preserve search term in session when user navigates between the pages.


mhigbyflowers (approved):
This would just remove the search term if reloaded?
Great solution!

PR #298, opened 2019-04-29
[risk=no] dealing with null exception on recursive tree node

SrushtiGangireddy:
1. Dealing with null exception on recursive tree node.


mhigbyflowers (approved):
👍 thanks

PR #305, opened 2019-04-30
[risk=no] prefix event category to differentiate from rh events

SrushtiGangireddy:
1. Prefix databrowser event category names to differentiate from research hub events.


mhigbyflowers (approved):
👍 thanks

PR #306, opened 2019-04-30
[DB-578] [risk=no] Survey chart conversion

SrushtiGangireddy:
1. Converting survey chart from pie to bar.
2. Adding missing age and gender deciles for survey results/


mhigbyflowers (approved):
Thanks looks good

PR #311, opened 2019-05-02
[DB-589] [risk=no] Changing other to display values <= 20 and no source concepts message

SrushtiGangireddy:
1. Other shows 0 if no data in there changing that to 20.
2. Added no source concepts message in there.


mhigbyflowers (commented):
👍 THanks!

mhigbyflowers (approved):
👍 Thanks!

PR #313, opened 2019-05-02
[risk=no] display only standard and source match pair for concept code / id match

SrushtiGangireddy:
1. Display only standard and source concept pair in the case of  concept id or code match.
2. Do not display name / synonym matched concepts.


mhigbyflowers (approved):
Thanks!

PR #323, opened 2019-05-06
[DB-520][risk=no] Log integration test failure logs

calbach:
also, compile integration tests on Circle


SrushtiGangireddy (approved):
Looks good. thanks.

calbach:
Interesting - I never actually committed the second part of this I said in my description.

PR #324, opened 2019-05-06
[DB-600] [risk=no] Add null values when unit is "no unit"

SrushtiGangireddy:
1. In the case of no unit, we do not have bin values, so just add the null values in there.


mhigbyflowers (approved):
👍 Thanks

PR #325, opened 2019-05-07
[DB-610] [risk=no] Grouping similar units

SrushtiGangireddy:
1. Created a manual map grouping similar units.
2. Using the destination unit in the map to create the counts.


mhigbyflowers (approved):
👍 looks good to me

PR #327, opened 2019-05-10
[DB-613] [risk=no] Call RolledupCountFetch api call only for snomed conditions / proocedures

SrushtiGangireddy:
1. RolledUpCriteria api was throwing errors on few measurements. We have not rolled up measurements yet so do not call api at all for other concepts.


mhigbyflowers (approved):
looks good, thanks

PR #328, opened 2019-05-10
[DB-612] [risk=no] deal with null pointer exception on search concepts api call

SrushtiGangireddy:
1. Trying to fetch content on empty concepts object was causing null pointer exception. Fixing that by adding a check in there.


mhigbyflowers (approved):
👍looks good

PR #337, opened 2019-05-14
[DB-626] [risk=no] investigate delay in drug search page load [revert-once-done]

SrushtiGangireddy:
1. Investigate why drug page is loading slow.


mhigbyflowers (approved):
👍

calbach (commented):
Are you able to reproduce the slowness locally? If so would be better to iterate there. Alternatively, if this needs to be on GAE you can push your own version up to test without merging this into master.

SrushtiGangireddy:
Ch, it is fine locally that's because there is no complete data.
And not able to push api changes from some time. 
So, planning to put this pr as a reference to revert the changes back.

calbach:
OK - not sure I follow what you mean on the API push, but for reference you can directly hit the deployed version even if it is not the primary promoted version:

https://drug-test003-dot-api-dot-aou-db-test.appspot.com/

You can also bring up a corresponding UI version to point to this if needed (though in this case, just `curl` on the API server is probably sufficient).

PR #338, opened 2019-05-14
[risk=no] changing data structure from tree set to array list

SrushtiGangireddy:
1. Method that adds missing bins to measurement concept results throws an error and many measurements do not show up in prod.
2. This was because tree set needs some sort of comparator which achilles result class do not have.
3. Using array list instead of tree set for now.

### DIFF
--- a/public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
+++ b/public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
@@ -1151,9 +1151,9 @@ public class DataBrowserController implements DataBrowserApiDelegate {
             }
         } else {
 
-            ArrayList<AchillesResult> maleResults = new ArrayList<AchillesResult>();
-            ArrayList<AchillesResult> femaleResults = new ArrayList<AchillesResult>();
-            ArrayList<AchillesResult> otherResults = new ArrayList<AchillesResult>();
+            List<AchillesResult> maleResults = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -1151,9 +1151,9 @@ else if(Integer.parseInt(ard.getStratum3()) == OTHER) {
             }
         } else {
 
-            TreeSet<AchillesResult> maleResults = new TreeSet<AchillesResult>();
-            TreeSet<AchillesResult> femaleResults = new TreeSet<AchillesResult>();
-            TreeSet<AchillesResult> otherResults = new TreeSet<AchillesResult>();
+            ArrayList<AchillesResult> maleResults = new ArrayList<AchillesResult>();
```
calbach: Better to use a general type if possible on the left-hand side, in this case `List`

calbach: In Java8 you can just exclude the generic type on the right-hand side `new ArrayList<>()`

calbach (commented):
Is this fixing a bug? Was this a performance issue?

calbach:
Oops - reread description, thanks. Sorry

SrushtiGangireddy:
Yup, it was throwing error and not opening up measurement value graphs in prod.

calbach (approved):

PR #343, opened 2019-05-16
[DB-626] [risk=low] Reducing load time by creating index on concept counts

SrushtiGangireddy:
1. Reduce the load time (especially of drug exposures) by reducing load time of drug concepts.
2. Index on count column would help anyways instead of having count > 0 or source count > 0 condition always.
3. Changed domain id param to string instead of list, since it is always passing only one domain.

### DIFF
--- a/public-api/db-cdr/changelog-schema/db.changelog-29.xml
+++ b/public-api/db-cdr/changelog-schema/db.changelog-29.xml
@@ -8,7 +8,6 @@
   <changeSet author="srushtigangireddy" runAlways="true" onValidationFail="MARK_RAN" id="changelog-29">
     <validCheckSum>ANY</validCheckSum>
     <addColumn tableName="concept">
-      <!-- Contains the tree path of the concept -->
       <column name="has_counts" type="BOOL">
         <constraints nullable="true"/>
       </column>
--- a/public-api/db-cdr/changelog-schema/db.changelog-29.xml
+++ b/public-api/db-cdr/changelog-schema/db.changelog-29.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd
+        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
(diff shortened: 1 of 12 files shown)


public-api/db-cdr/changelog-schema/db.changelog-29.xml
```diff
@@ -0,0 +1,23 @@
...
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd
+        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+  <changeSet author="srushtigangireddy" runAlways="true" onValidationFail="MARK_RAN" id="changelog-29">
+    <validCheckSum>ANY</validCheckSum>
+    <addColumn tableName="concept">
+      <!-- Contains the tree path of the concept -->
```
calbach: stale?

public-api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -633,6 +634,16 @@ from (select analysis_id, concept_id, stratum_1 as stratum, domain, max(count_va
 group by analysis_id, concept_id, stratum, domain) as sub_cr
 where cast(sub_cr.concept_id as string)=c.stratum_1 and c.analysis_id=sub_cr.analysis_id and c.stratum_2=cast(sub_cr.stratum as string) and c.stratum_3=sub_cr.domain"
 
+bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
+"Update \`$OUTPUT_PROJECT.$OUTPUT_DATASET.concept\` c
+set has_counts = 1
```
calbach: One command possible here? `set has_counts = (c.count_value > 0 or c.source_count_value > 0)`

public-api/src/main/resources/public-api.yaml
```diff
@@ -339,6 +339,10 @@ definitions:
         type: "array"
         items:
           type: "string"
+      hasCounts:
```
calbach: Do we need this in the API? seems like it should be easy for the client to check this for themselves using the current API model.

If not, I would suggest just keeping it an internal implementation detail

public-api/db-cdr/changelog-schema/db.changelog-29.xml
SrushtiGangireddy: Done thanks.

public-api/db-cdr/generate-cdr/make-bq-data.sh
SrushtiGangireddy: Done thanks.

public-api/src/main/resources/public-api.yaml
SrushtiGangireddy: Added it in there for tests to run, otherwise was not using it anywhere. Fixed it. Thanks.

calbach (approved):

PR #346, opened 2019-05-20
[DB-595] [risk=no] Overflow tooltips when long

SrushtiGangireddy:
1. Overflow tooltips when the length of labels is long.


mhigbyflowers (commented):
👍 thanks

mhigbyflowers (approved):
👍

PR #348, opened 2019-05-20
[DB-505] [risk=no] Align synonyms text with the icons on top 

SrushtiGangireddy:
Align the text in concept synonyms when clicked in see more with the text on the top.


mhigbyflowers (approved):
Looks good

PR #351, opened 2019-05-23
[DB-635] [risk=no] Adding other in pm value graphs and rounding missing values to 2 decimals

SrushtiGangireddy:
1. Round the missing values to 2 decimals.
2. Add total before overall participant count on pregnancy and wheel chair concepts.
3. Add other value graph display.


mhigbyflowers (approved):
👍 Looks good

PR #352, opened 2019-05-30
[DB-592] [risk=no] use source count value in sources bar chart

SrushtiGangireddy:
1. Use source count value in the place of count value in sources bar chart.

### DIFF
--- a/public-ui/src/app/data-browser/chart/chart.component.ts
+++ b/public-ui/src/app/data-browser/chart/chart.component.ts
@@ -338,7 +338,6 @@ export class ChartComponent implements OnChanges, AfterViewInit {
   }
 
   public makeConceptChartOptions() {
-    console.log('here');
     const data = [];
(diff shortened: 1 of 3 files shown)


public-ui/src/app/data-browser/chart/chart.component.ts
```diff
@@ -338,29 +338,46 @@ export class ChartComponent implements OnChanges, AfterViewInit {
   }
 
   public makeConceptChartOptions() {
+    console.log('here');
```
mhigbyflowers: delete this log

mhigbyflowers (approved):
👍, see comment

public-ui/src/app/data-browser/chart/chart.component.ts
SrushtiGangireddy: Done thanks.

PR #353, opened 2019-05-30
[DB-582] [risk=no] Remove the focus on chart and make first chart type default for all domains

SrushtiGangireddy:
1. Remove the focus on the chart and make values in measurement default opening tab. (First)


mhigbyflowers (approved):
👍  looks good. Thanks

PR #354, opened 2019-05-30
[DB-511] Making the bucket ranges more readable

SrushtiGangireddy:
1. Smoothing the distribution curve and making the value bucket range multiples of 10.

### DIFF
--- a/public-ui/src/app/views/ehr-view/ehr-view.component.ts
+++ b/public-ui/src/app/views/ehr-view/ehr-view.component.ts
@@ -58,9 +58,6 @@ export class EhrViewComponent implements OnInit, OnDestroy {
   treeLoading = false;
   search: string;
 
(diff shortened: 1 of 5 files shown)


public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -238,8 +235,6 @@ export class EhrViewComponent implements OnInit, OnDestroy {
   }
 
   public selectGraph(g, r: any) {
-    this.chartEl.nativeElement.scrollIntoView(
-      { behavior: 'smooth', block: 'nearest', inline: 'start' });
```
mhigbyflowers, line 242: 🙌

mhigbyflowers (approved):
👍

PR #359, opened 2019-06-03
[risk=no] do not display mix of text and numeric values

SrushtiGangireddy:
1. Check pre-conditions before adding column.
2. Do not display numeric values (in case of no unit) if there are text values. That mix up looks bad.


mhigbyflowers (approved):
👍 thanks

PR #362, opened 2019-06-05
[DB-662] [risk=no] measurement bin updates

SrushtiGangireddy:
1. Make the bin range of all the genders same for any measurement.
2. Round bin width to one decimal place.
3. Generate physical measurements participant counts
4. Rounding the min and max to multiples of 5 of range from 32 to 30 for readbility.


mhigbyflowers (approved):
👍

PR #365, opened 2019-06-07
[DB-658] [risk=no] Fix the physical measurement card to look like any other domain card

SrushtiGangireddy:
1. Add participant count to pm card.


mhigbyflowers (approved):
👍 thanks

PR #366, opened 2019-06-11
[DB-607] [risk=no] change refetch time on search

SrushtiGangireddy:
1. Change the search call back time to reduce the number of api calls.


mhigbyflowers (approved):
👍 looks good

PR #371, opened 2019-06-17
[DB-669] [risk=low] Proper binning physical measurements

SrushtiGangireddy:
1. Binning physical measurements correctly.
2. Displaying pregnancy, wheel chair user stacked biological sexes bars in specific order.


mhigbyflowers (approved):
👍 looks good

PR #375, opened 2019-06-20
[risk=no] [DB-679] Add physical measurements results when no results in other domains

SrushtiGangireddy:
1. When no results in other domain and physical measurements has relevant results display the link.


mhigbyflowers (approved):
👍 thanks

PR #376, opened 2019-06-20
[risk=no] [DB-674] 2 letter concept code match bug

SrushtiGangireddy:
1. Concept code match does not work on two letter search term


mhigbyflowers (approved):
👍 Looks good

PR #381, opened 2019-06-25
[risk=no] [DB-682] fix the second level survey sub question's did not answer count

SrushtiGangireddy:
1. Use the count of answer from which the second level sub question branched to calculate did not answer count instead of count of question.


mhigbyflowers (approved):
👍

PR #384, opened 2019-06-25
[risk=no] [DB-587] Update how sub question count % is calculated

SrushtiGangireddy:
1. Update the calculation to be: (# participants who selected that sub-answer / # of participants who answered *that su‌‌‌‌‌bquestion*).


mhigbyflowers (approved):
Looks good. Thanks

PR #388, opened 2019-06-25
[rsik=no] cdr version update in stable and staging

SrushtiGangireddy:
1. Updating cdr version in stable and staging


mhigbyflowers (approved):
👍

PR #392, opened 2019-07-03
[DB-688] [risk=no] Put measurement data in proper bins

SrushtiGangireddy:
1. Mix and match a little to make measurement bins more readable.


mhigbyflowers (approved):
Looks good 👍

PR #394, opened 2019-07-05
[risk=no] [DB-649] bold the answer to know which survey sub question is opened up

SrushtiGangireddy:
1. Bold the survey answer clicked / expanded that shows the next level survey question for user-readability.


mhigbyflowers (approved):
looks good 👍

PR #396, opened 2019-07-15
[risk=no] [DB-720] Bucket more than one selected race / ethnicity

SrushtiGangireddy:
1. Bucket more than one selected race / ethnicity into different bucket.
2. Do not count multiple answers in rest of teh r / e buckets.


mhigbyflowers (approved):
👍

PR #402, opened 2019-07-23
[risk=low] [DB-728] Do not display range for no unit measurement values

SrushtiGangireddy:
1. Measurement values in no-unit displays range like the range, which does not make sense.
2. Code change to accommodate better view.


mhigbyflowers (approved):
👍 thanks

PR #404, opened 2019-07-25
[risk=no[ [DB-748] fixing duplication in survey charts

SrushtiGangireddy:
1. Remove duplicate biological sex and age bars in charts that came from the other path of same sub question.


mhigbyflowers (approved):
👍  looks good

PR #407, opened 2019-08-07
[DB-684] show percentage, count buttons for only gender and age charts

SrushtiGangireddy:
1. Fix the buttons to show up only for gender and age charts.


mhigbyflowers (approved):
👍 thanks

PR #409, opened 2019-08-08
[DB-774] do not modify search keyword if it has quotes in there

SrushtiGangireddy:
1. Search seems to be broken when search keyword has double quotes around it.


mhigbyflowers (approved):
👍

PR #411, opened 2019-08-12
[DB-747] Replace greater than or equal to symbol in chart labels

SrushtiGangireddy:
1. Replace >= with actual symbol.


mhigbyflowers (approved):
👍 thanks

mhigbyflowers (approved):
looks good 👍

PR #412, opened 2019-08-12
[DB-764] fixing tooltip on sources chart

SrushtiGangireddy:
The count on tooltip on sources chart does not match the actual count. 
Fixing this bug.


mhigbyflowers (approved):
👍 looks good, thanks

PR #415, opened 2019-08-15
[risk=no] [DB-743] do not display values in measurements that are greater than max

SrushtiGangireddy:
1. Charts show weird bucket range.
2. Do not display the range > max bucket value.


mhigbyflowers (approved):
👍  thanks

PR #416, opened 2019-08-21
[risk=low] [DB-780] Check the percentage for rare denominators

SrushtiGangireddy:
1. In the case of adding missing biological sex / age results with 20 counts. 
Calculate denominator with count 20 and not 0.
2. Little tweaks in measurement queries.


mhigbyflowers (approved):
looks good 👍

PR #417, opened 2019-08-26
[DB-781] [risk=no] Flip Count and percentage charts order

SrushtiGangireddy:
1. Show count chart on default


mhigbyflowers (approved):
👍 looks good

PR #418, opened 2019-08-27
[DB-775] [risk=no] configure gtm events for count / percentage tab clicks

SrushtiGangireddy:
1. Configure events for count / percentage (%) tab clicks


mhigbyflowers (approved):
👍  looks good 
see conflict:

PR #420, opened 2019-09-03
[risk=no] [DB-723] Tag more than one race/ethnicity option

SrushtiGangireddy:
1. Add tagging to more than one race / ethnicity option.


mhigbyflowers (approved):
👍

PR #422, opened 2019-09-04
[risk=no] [DB-788] deal with having / in search term

SrushtiGangireddy:
1. Having '/' in search term was breaking the search as there was no matching route


SrushtiGangireddy:
@mhigbyflowers Let me know if you think of a better way to solve this.

mhigbyflowers (approved):
Looks good 👍 thanks

PR #423, opened 2019-09-04
[risk=no] [DB-789] do not display full label on yaxis if the text value is too big

SrushtiGangireddy:
1. Graph looks weird with long yaxis text labels.


mhigbyflowers (approved):
👍 thanks

PR #426, opened 2019-09-10
[risk=no] measurement query fix

SrushtiGangireddy:
1. Fixing query to fix few measurement value distributions.


mhigbyflowers (approved):
👍

PR #429, opened 2019-09-16
[risk=no] [DB-814] save drug results and display them over pages

SrushtiGangireddy:
1. Drug results are extra and are messing up with pagination.
2. Do not fetch drug results in subsequent page. (Because we would already have them with 1st page search).
3. Save them and display over pages.


mhigbyflowers (approved):
thanks for catching this!

PR #437, opened 2019-09-19
[risk=no] [DB-699] back button fix when no search term

SrushtiGangireddy:
Back button not working when no search term


mhigbyflowers (approved):
👍

PR #438, opened 2019-09-19
[risk=no] [DB-699] Fix back button

SrushtiGangireddy:
Fix back button working by adding replaceUrl param to router navigate event to not override browser history (which was causing the control to be in the same page infinitely)


mhigbyflowers (approved):
Looks good, Thanks 👍

PR #445, opened 2019-10-01
[risk=no] fix percentage charts tooltip and concept code match display

SrushtiGangireddy:
1. Fix percentage chart tooltip and concept code match display.


mhigbyflowers (approved):
👍 thanks

PR #451, opened 2019-10-10
[risk=no] Font awesome link fix and speed up domain results

SrushtiGangireddy:
1. Having 2 dao calls in domain info slowed the domain totals api. Fixed that to be one dao call instead like before.


mhigbyflowers (approved):
👍 looks good

PR #453, opened 2019-10-11
[risk=no] [DB-806] fixing ranges in measurements

SrushtiGangireddy:
1. Some value ranges were missing in measurements. Fixing them.


mhigbyflowers (approved):
👍

PR #458, opened 2019-10-24
[risk=no] Map survey question text to match what's in the pdf

SrushtiGangireddy:
1. Map the survey question text to match what is in the pdf


mhigbyflowers (approved):
thanks for all this work 👍

PR #460, opened 2019-10-28
Revert "[risk=no] Map survey question text to match what's in the pdf"

SrushtiGangireddy:
Reverts all-of-us/data-browser#458


mhigbyflowers (approved):
👍

PR #470, opened 2019-11-07
[risk=no] speed up sources api

SrushtiGangireddy:
1. Added index only on parent_id in cb_criteria.
2. Removing unused children from CriteriaParentResponse.


mhigbyflowers (approved):
Thanks!

PR #479, opened 2019-11-11
[risk=no] [DB-834] removing ai an from other tooltip

SrushtiGangireddy:
1. Removing ai an from other tooltip


mhigbyflowers (approved):
👍

PR #484, opened 2019-11-15
[risk=no] add Flag to concept object

SrushtiGangireddy:
1. Add flag to concept object
2. Pass search text value to domain search results api.


mhigbyflowers (approved):
Thanks 👍

PR #487, opened 2019-11-18
[risk=no] Do not load concept graph twice

SrushtiGangireddy:
1. Setting search text value in loadPage function is triggering the search text value change and api request fire which is causing reload.


public-ui/src/app/data-browser/views/ehr-view/ehr-view.component.ts
```diff
@@ -134,7 +134,7 @@ export class EhrViewComponent implements OnInit, OnDestroy {
         this.prevSearchText = localStorage.getItem('searchText');
       }
     }
-    this.searchText.setValue(this.prevSearchText);
+    // this.searchText.setValue(this.prevSearchText);
```
mhigbyflowers: Thanks for finding this.
I think we can just remove this line.

mhigbyflowers (approved):
👍  thanks!

PR #492, opened 2019-11-20
[risk=no][DB-849]adding loadSourceTree to expandRow 

mhigbyflowers:
this fixes the issue when a user clicks the top chart they would get the previous source tree.


SrushtiGangireddy (approved):
Thanks

PR #498, opened 2019-12-05
[DB-767] [risk=no] show drug brand names for each ingredient

SrushtiGangireddy:
1. Show drug brand name for each ingredient.


mhigbyflowers (approved):
looks good 👍

PR #500, opened 2019-12-06
[risk=no] [DB-876] Weird body weight values

SrushtiGangireddy:
map source to standard unit concepts and do not use source units anymore


mhigbyflowers (approved):
👍

PR #508, opened 2019-12-16
[risk=no] [DB-881] fixing bug in manual question map

SrushtiGangireddy:
1. Fixing bug in manual survey question map to fix double display issue of one question.


mhigbyflowers (approved):
👍

PR #510, opened 2019-12-17
Matching Alpha Header[db-869][risk=no] 

mhigbyflowers:
matching the alpha header for the research hub.
• created an environment variable called isAlpha that when true, the "Apply" and "Research Login" button will populate


public-ui/src/environments/test-env-base.ts
```diff
@@ -5,6 +5,6 @@ export const testEnvironmentBase = {
   publicApiUrl: 'https://api-dot-aou-db-test.appspot.com',
   workbenchUrl: 'https://all-of-us-workbench-test.appspot.com',
   researchAllOfUsUrl: 'https://staging.researchallofus.org',
-  isAlpha: true,
+  isAlpha: false,
```
mhigbyflowers, line 8: We will set this to true when for the Alpha launch

SrushtiGangireddy (approved):
Looks good thanks

PR #523, opened 2019-12-24
[risk=no] [DB-885] let sources graph take full space in drugs / measurements 

SrushtiGangireddy:
let sources graph take full space in drugs / measurements as no sources tree yet


mhigbyflowers (approved):
👍  looks good. 
I took out some unneeded template stuff

PR #524, opened 2019-12-24
[risk=no] [DB-886] Update <= 20 on y axis

SrushtiGangireddy:
1. Replace <= 20 on y axis when any of the bars has value > 20. Start y axis from 0 instead.


SrushtiGangireddy:
image

SrushtiGangireddy:
image

mhigbyflowers (approved):
looks good, ... lets circle back with Helen because I don't think we are supposed to show zeros

PR #525, opened 2019-12-24
[risk=no] [DB-715] Check and adjust text contrast ratio

SrushtiGangireddy:
1.Adjusting colors and font-size of text to be complaint with text contrast ratio


mhigbyflowers (commented):
Looks good to me, but let's hold off and circle back with Helen

PR #530, opened 2019-12-30
[risk=no] [DB-887] Weird no-unit values in measurement graphs

SrushtiGangireddy:
1. Unit source value is sometimes '' which we consider as no-unit so add this check to condition.


mhigbyflowers (approved):
looks good, thanks

PR #534, opened 2020-01-02
[risk=no] changing style of drug aka text

SrushtiGangireddy:
1. Making see more button more prominent and style change for it to not seem to be for aka-text.


mhigbyflowers (approved):
looks good 👍 thanks

PR #535, opened 2020-01-03
[DB-765][risk=no] new footer

mhigbyflowers:
looks like a pulled in the color and font adjustment by mistake, but we are going forward with it so I think it should be fine.


SrushtiGangireddy (approved):
Looks good except for the test fail in app.component.html, thanks.

PR #541, opened 2020-01-09
[risk=no] [DB-898] roll up icd10 condition counts

SrushtiGangireddy:
1. Roll up icd10 condition concepts


mhigbyflowers (approved):
👍 look good, thanks

PR #542, opened 2020-01-14
[risk=no] [DB-892] restructure and speed up survey

SrushtiGangireddy:
1. Splitting the api call as huge data fetch causing the api to be slow.


mhigbyflowers (approved):
thanks,
looks good 👍

PR #546, opened 2020-01-22
[risk=no] [DB-892] Font sizes update

SrushtiGangireddy:
try making the smallest to a size 14 (graph labels, tooltips). the rest to a size 18

See if we can update tooltips' color to be white with dark purple font. Might need a shadow to make it pop off the page? Try the shadow like on the homepage tiles. Let’s try font size 18, maybe 14.

### DIFF
--- a/common-api/src/main/java/org/pmiops/workbench/cdr/dao/SurveyModuleDao.java
+++ b/common-api/src/main/java/org/pmiops/workbench/cdr/dao/SurveyModuleDao.java
@@ -13,21 +13,15 @@ public interface SurveyModuleDao extends CrudRepository<SurveyModule, Long> {
    * @param matchExpression a boolean full text match expression based on the user's query; see
    *                https://dev.mysql.com/doc/refman/5.7/en/fulltext-boolean.html
    */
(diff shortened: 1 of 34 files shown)


mhigbyflowers (commented):
image
This tooltip is missing

mhigbyflowers (changes requested):
image
This tooltip is missing

The font changes on the percent and count look good. I think we should push those changes and wait to push the tooltip background changes.

SrushtiGangireddy:
image

SrushtiGangireddy:
image

SrushtiGangireddy:
Looks okay to me. It needs local data migrations runs may be.
(since we changed survey name from healthcare to "health care")

mhigbyflowers (approved):
thanks for fixing. 👍 
I think the black tool-tip looks better, but I know 508 says it is more assessable.  Lets merge it in to test and see what Hellen thinks.

SrushtiGangireddy:
Yes, i was just going by ticket. Let us see. We can revert back from test.

PR #548, opened 2020-01-24
[risk=no] changing test cdr version

SrushtiGangireddy:
1. Changing test cdr version.
2. Added new indexes on survey question map table


mhigbyflowers (approved):
👍

PR #571, opened 2020-03-09
[risk=no] [DB-967] changing tooltip on fmh page

SrushtiGangireddy:
Family medical history component is still using old tooltip.
Changing it.


mhigbyflowers (approved):
👍

PR #572, opened 2020-03-10
workBenchIsBeta feature flag[risk=no]

mhigbyflowers:
sets this flag to true so the header and footer will match rh.org 
DO NOT MERGE until we are ready for beta


mhigbyflowers:
Not sure why the API is failing.

SrushtiGangireddy:
Should we also change this flag in other environments ?

SrushtiGangireddy (approved):
Looks good, thanks.

PR #574, opened 2020-03-16
[risk=no][DB-867]Keyboard navigation for the header

mhigbyflowers:
this PR is to allow tabbing through the header and navigating links with the keyboard

### DIFF
--- a/public-ui/src/app/shared/components/rh-header/rh-header.component.css
+++ b/public-ui/src/app/shared/components/rh-header/rh-header.component.css
@@ -254,7 +254,8 @@ li.flip::before, li.flip::before {
 .site-nav-container nav.main-navigation .menu-primary-container>ul.main-menu>li>ul.sub-menu {
   list-style: none;
   padding-left: 1rem;
-  display: none;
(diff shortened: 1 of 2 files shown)


public-ui/src/app/shared/components/rh-header/rh-header.component.css
```diff
@@ -243,18 +241,17 @@ li.flip::before, li.flip::before {
...
+.site-nav-container nav.main-navigation .menu-primary-container ul.main-menu li ul.sub-menu, .sub-menu li ul {
   list-style: none;
 }
 
 .site-nav-container nav.main-navigation .menu-primary-container>ul.main-menu>li>ul.sub-menu {
   list-style: none;
   padding-left: 1rem;
-  display: none;
+  top: -10000px;
+  position: absolute;
```
mhigbyflowers, line 254: the first state of the dropdown menu is off the page. This allows the links to be tabbed through because it never actually disappears from the DOM.

SrushtiGangireddy (approved):
Looks good. Thanks.

PR #575, opened 2020-03-19
[risk=no] [DB-631] tearing down backend sign-in code

SrushtiGangireddy:
Tearing down sign in code on backend.


common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -11,7 +11,6 @@
 
   public static class AuthConfig {
```
calbach: I think this whole section should be removed

calbach (approved):
Looks good once all the configs have been updated as well  (test failures).

SrushtiGangireddy:
Thanks.

PR #604, opened 2020-07-16
[risk=no] [DB-983] adding frame headers

SrushtiGangireddy:
Adding frame headers to secure the app from clickjacking.


mhigbyflowers (approved):
great Thanks 👍

PR #605, opened 2020-07-20
[risk=no] [DB-619] Physical Measurement height display values in inches too

SrushtiGangireddy:
Add height values also in american default unit inches.


mhigbyflowers (approved):
Looks good to me 👍

PR #607, opened 2020-07-31
[risk=no] [DB-947] Use prod data in test/ staging/ stable

SrushtiGangireddy:
Use prod data in test/ staging and stable environments

### DIFF
--- a/public-api/db-cdr/generate-cdr/move-aggregated-counts.sh
+++ b/public-api/db-cdr/generate-cdr/move-aggregated-counts.sh
@@ -6,19 +6,19 @@
 # 0) .csv of prod tables in a bucket
 
 # Example usage, you need to provide a bunch of args
-# ./project.rb move-aggregated-counts --prod-project aou-db-prod --public-project aou-db-test \
+# ./project.rb move-aggregated-counts --prod-project aou-db-prod --test-project aou-db-test \
 # --prod-bucket aou-db-prod-public-cloudsql --test-bucket aou-db-public-cloudsql --cdr-version p_2020q2_1
 
 set -xeuo pipefail
 IFS=$'\n\t'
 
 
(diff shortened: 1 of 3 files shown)


public-api/db-cdr/generate-cdr/move-aggregated-counts.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# This moves aggregated count data csvs from prod bucket to public bucket to be imported to cloudsql
```
calbach: Is public being used here to indicate an actual public ACL? Or is this using public in the way that databrowser normally uses public (e.g. as a database name).

In practice, what is the destination project used for --public-project ?

SrushtiGangireddy: hmmm, it means public aka "test" bucket/ project.

calbach: OK. Should it be called "test-bucket" perhaps? Or something else? I'm just getting confused as to whether this means the ACL on this bucket actually public - is it?

SrushtiGangireddy: Ch, changed the wording to be test-bucket. The bucket is not public to all users though. It has access to few service accounts and our team.

public-api/db-cdr/generate-cdr/move-aggregated-counts.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# This moves aggregated count data csvs from prod bucket to test bucket to be imported to cloudsql
+
+# End product is:
+# 0) .csv of prod tables in a bucket
+
+# Example usage, you need to provide a bunch of args
+# ./project.rb move-aggregated-counts --prod-project aou-db-prod --public-project aou-db-test \
```
calbach: would probably change `--public-project` --> `--test-project` for consistency

calbach (approved):

PR #610, opened 2020-08-06
[risk=low] setting up iap on test

SrushtiGangireddy:
Since we are now showing prod data in test, setting up iap on test environment.

### DIFF
--- a/public-ui/dispatch.yaml
+++ b/public-ui/dispatch.yaml
@@ -1,0 +1,7 @@
+dispatch:
+  - url: "*/v1/*"
+    service: api
+  - url: "*/favicon.ico"
+    service: default
+  - url: "aou-db-test.appspot.com/"
+    service: default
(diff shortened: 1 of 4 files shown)


public-ui/dispatch.yaml
```diff
@@ -0,0 +1,7 @@
+dispatch:
+  - url: "*/v1/*"
```
calbach: This seems slightly too broad, e.g. if there were a UI page which had a "/v1/" somewhere in the path. Can you restrict it to "/v1/*", or can you inclue the base uRL in here?

SrushtiGangireddy: added base url. thanks.

public-ui/dispatch.yaml
```diff
@@ -0,0 +1,7 @@
+dispatch:
+  - url: "aou-db-test.appspot.com/v1/*"
```
calbach, line 2: This looks fine as long as higher environments can ignore this. I guess just double-check after the next staging push.

calbach (approved):

PR #617, opened 2020-08-17
[DB-632][risk=no] fixed EHR header

mhigbyflowers:
allows the header to be fixed in the EHR view


SrushtiGangireddy (approved):
Looks good thanks.

PR #635, opened 2020-09-09
[risk=no] Authenticate integration test api calls on staging

SrushtiGangireddy:
Integration tests fail on the staging environment with forbidden error.
Using authentication with bearer token to get past iap.


public-api/src/main/java/org/pmiops/workbench/google/IAPAuthService.java
```diff
@@ -0,0 +1,110 @@
...
+                .buildPostRequest(new GenericUrl(OAUTH_TOKEN_URI), content)
+                .setParser(new JsonObjectParser(JacksonFactory.getDefaultInstance()));
+
+        HttpResponse response = request.execute();
+        GenericData responseData = response.parseAs(GenericData.class);
+        String idToken = (String) responseData.get("id_token");
+        return JWT.decode(idToken);
+    }
+
+    private String getSignedJwt() {
```
calbach: Is it not possible to use this code snippet from the documentation, instead of the explicit JWT signing? https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/iap/src/main/java/com/example/iap/BuildIapRequest.java

public-api/src/main/java/org/pmiops/workbench/google/IAPAuthService.java
```diff
@@ -0,0 +1,110 @@
...
+    public String getToken() {
+        return token;
+    }
+
+    public void setToken(String token) {
+        this.token = token;
+    }
+
+    private ServiceAccountCredentials getCredentials() throws IOException {
+        Resource resource = resourceLoader.getResource("classpath:test-circle-key.json");
```
calbach: It should be possible to use application default credentials here ( `GoogleCredentials.getApplicationDefault().createScoped(Collections.singleton(IAM_SCOPE))`), if you wrap the integration test launch in a service account context, see: https://github.com/all-of-us/workbench/blob/899ad3051f3c812234e22f498f3ae24d2f42e758/api/libproject/devstart.rb#L430

This would avoid needing to hardcode the resource path here

public-api/src/integration/java/org/pmiops/workbench/publicapi/DataBrowserControllerIntegrationTest.java
```diff
@@ -38,11 +39,23 @@
     @Bean
     DataBrowserApi client() {
       DataBrowserApi api = new DataBrowserApi();
+      IAPAuthService iapAuth;
+      String token = "";
+      try {
+        iapAuth = new IAPAuthentication();
+        token = iapAuth.getToken();
+      } catch (IOException ie) {
```
calbach: just throw from the provider method

public-api/src/main/java/org/pmiops/workbench/google/IAPAuthService.java
SrushtiGangireddy: Changed it thanks.

public-api/libproject/devstart.rb
```diff
@@ -367,6 +375,13 @@ def run_integration_tests(cmd_name, *args)
   :fn => ->(*args) { run_integration_tests("integration", *args) }
 })
 
+def get_gsuite_admin_key(project)
```
calbach: data browser should not have a gsuite admin key

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
```diff
@@ -0,0 +1,64 @@
...
+    public static void buildIapRequest(String iapClientId)
+            throws IOException {
+
+        IdTokenProvider idTokenProvider = getIdTokenProvider();
+        IdTokenCredentials tokenCredential = IdTokenCredentials.newBuilder()
+            .setIdTokenProvider(idTokenProvider)
+            .setTargetAudience(iapClientId).build();
+        Map<String, List<String>> requestData = tokenCredential.getRequestMetadata();
+        if (requestData.containsKey("Authorization")) {
+            TOKEN = requestData.get("Authorization").get(0);
```
SrushtiGangireddy: @calbach getIdToken() and getAccessToken() on tokenCredential are somehow returning null. 
The authorization token seems to be in request metadata.

calbach: weird, perhaps try calling `refresh()` first

public-api/libproject/devstart.rb
SrushtiGangireddy: Cleaned it up. Than ks.

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
SrushtiGangireddy: Thanks. Worked after calling refresh.

public-api/libproject/devstart.rb
```diff
@@ -147,33 +148,34 @@ def read_db_vars(gcc)
...
     raise("Please run 'gcloud auth login' before starting the server.")
   end
 
-  at_exit { common.run_inline %W{docker-compose down} }
-  common.status "Starting database..."
-  common.run_inline %W{docker-compose up -d db}
-  common.status "Running database migrations..."
-  common.run_inline %W{docker-compose run db-migration}
-  common.run_inline %W{docker-compose run db-public-migration}
+  ServiceAccountContext.new(TEST_PROJECT).run do
```
calbach: This should not be necessary. THis may have been from when you were having issues with ADC from the server.

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.auth.http.HttpCredentialsAdapter;
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.IdTokenCredentials;
+import com.google.auth.oauth2.IdTokenProvider;
+import com.google.common.base.Preconditions;
+import java.io.IOException;
+import java.util.*;
+
+public class BuildIapRequest {
```
calbach, line 17: nit: please link back to the documentation where this came from

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
```diff
@@ -0,0 +1,62 @@
...
+    private static final String IAM_SCOPE = "https://www.googleapis.com/auth/iam";
+
+    private static final HttpTransport httpTransport = new NetHttpTransport();
```
calbach: unused

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private static String TOKEN = "";
```
calbach: I'm not sure if this was from experimentation, but the static usage here is a bit sketchy. I think you really just want a single method on the class that fetches a token.

public-api/src/integration/java/org/pmiops/workbench/publicapi/DataBrowserControllerIntegrationTest.java
```diff
@@ -32,17 +34,27 @@
 public class DataBrowserControllerIntegrationTest {
 
   public static String DB_API_BASE_PATH = "DB_API_BASE_PATH";
+  public static String CLIENT_ID = "238501349883-965gu9qminos5dfcpusi43eokvd5i3io.apps.googleusercontent.com";
```
calbach, line 38: This is only the staging client ID right? If so, should clarify this, or else make this an environment variable provided to the test.

public-api/libproject/devstart.rb
SrushtiGangireddy: Thanks removed.

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
SrushtiGangireddy: Done thanks.

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
SrushtiGangireddy: Removed thanks.

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
SrushtiGangireddy: Removed thanks.

public-api/src/integration/java/org/pmiops/workbench/publicapi/DataBrowserControllerIntegrationTest.java
SrushtiGangireddy: Added a comment on top.

public-api/libproject/devstart.rb
```diff
@@ -341,6 +342,7 @@ def run_all_tests(cmd_name, args)
 
 def run_integration_tests(cmd_name, *args)
   ensure_docker cmd_name, args
+  common = Common.new
```
calbach: nit: rm, this is defined later on as well

public-api/build.gradle
```diff
@@ -159,7 +159,7 @@ task integration(type: Test) {
   testClassesDirs = sourceSets.integration.output.classesDirs
   classpath = sourceSets.integration.runtimeClasspath
   systemProperties = [
-    'DB_API_BASE_PATH': System.getenv('DB_API_BASE_PATH')
+          'DB_API_BASE_PATH': System.getenv('DB_API_BASE_PATH')
```
calbach: nit: revert

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.*;
+
+/**
+ Class that can be used to get OIDC token for a service account.
+ https://cloud.google.com/iap/docs/authentication-howto#iap_make_request-java
+ */
+public class BuildIapRequest {
+    private static final String IAM_SCOPE = "https://www.googleapis.com/auth/iam";
+
+    public BuildIapRequest() throws IOException {
```
calbach: doesn't throw

calbach: opt: Probably just use a private constructor in this case, since it's not meant to be instantiated

public-api/src/integration/java/org/pmiops/workbench/publicapi/DataBrowserControllerIntegrationTest.java
```diff
@@ -42,7 +46,18 @@ DataBrowserApi client() {
...
       }
-      api.setApiClient(new ApiClient().setBasePath(basePath));
+      ApiClient apiClient = new ApiClient();
+      apiClient.setBasePath(basePath);
+      if (basePath.contains("aou-db-staging")) {
+        String token = "";
+        try {
+          BuildIapRequest buildRequest = new BuildIapRequest();
+          token = buildRequest.buildIapRequest(CLIENT_ID);
+        } catch (IOException ie) {
```
calbach: should probably just let this go uncaught so the test fails

calbach (approved):
Sorry, I forgot you were probably waiting for me on this. Looks good, but please make the remaining requested changes.

public-api/libproject/devstart.rb
SrushtiGangireddy: Done thanks.

public-api/build.gradle
SrushtiGangireddy: Done.

public-api/src/integration/java/org/pmiops/workbench/google/BuildIapRequest.java
SrushtiGangireddy: Removed constructor. Thought it will. use java's default constructor.

public-api/src/integration/java/org/pmiops/workbench/publicapi/DataBrowserControllerIntegrationTest.java
SrushtiGangireddy: I cannot not catch it. I should either catch it or throw it.

calbach: You should be able to simply change the method signature to `DataBrowserApi client() throws IOException {`

SrushtiGangireddy: Throwing from client has thrown error previously for a different reason may be. That whole branch was messed up. It works fine now. Thanks.

SrushtiGangireddy:
Thanks.

PR #644, opened 2020-09-21
[risk=no] adding check to add SA context

SrushtiGangireddy:
Adding check to test for aou-db-staging env and add SA context.

### DIFF
--- a/public-api/libproject/devstart.rb
+++ b/public-api/libproject/devstart.rb
@@ -359,7 +359,12 @@ def run_integration_tests(cmd_name, *args)
 
   common = Common.new
   common.status "Executing integration tests against '#{api_base}'"
-  ServiceAccountContext.new(TEST_PROJECT).run do
+
+  if op.opts.env == 'aou-db-staging'
+    ServiceAccountContext.new(TEST_PROJECT).run do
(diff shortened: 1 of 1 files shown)


public-api/libproject/devstart.rb
```diff
@@ -359,7 +359,12 @@ def run_integration_tests(cmd_name, *args)
 
   common = Common.new
   common.status "Executing integration tests against '#{api_base}'"
-  ServiceAccountContext.new(TEST_PROJECT).run do
+
+  if op.opts.env == 'aou-db-staging'
+    ServiceAccountContext.new(TEST_PROJECT).run do
```
calbach: Please add a comment here explaining why it's only necessary for staging, also please verify this works when you run it locally `./project integration --project aou-db-staging`

Would also like to understanding whyt he appspot aou-db-test is being used here. Seems like it should probably be  the circle-deploy-account

calbach (approved):

public-api/libproject/devstart.rb
SrushtiGangireddy: Thanks. Added a comment in there. This works fine locally.

PR #645, opened 2020-09-24
[no-ticket]adding ts-fetch codegen

mhigbyflowers:
@SrushtiGangireddy : you might need to `git pull origin master` within `./aou-utils` because I updated the codegen.jar in the aou-utils repo  <------- nope looks like I was able to commit that change.


SrushtiGangireddy (approved):
LGTM. Thanks.

PR #657, opened 2020-10-09
[risk=no] Cope data counts

SrushtiGangireddy:
1. New chart data for count by versions.
2. Added stratum 7 in 3110 rows to hold flag that indicates if it has a sub - question.


mhigbyflowers (approved):
LGTM

PR #660, opened 2020-10-14
[risk=no] Changing api call to speed up survey

SrushtiGangireddy:
Drilling down api calls instead of caching everything at same time.


public-ui/src/app/data-browser/views/quick-search/quick-search.component.ts
```diff
@@ -232,7 +232,7 @@ export class QuickSearchComponent implements OnInit, OnDestroy {
     this.surveyResults = results.surveyModules;
     // TODO remove this filter when the feature flag is turned off or to debug
     if (!this.cope) {
-      this.surveyResults = results.surveyModules.filter(sm => sm.conceptId !== 1333342);
+      // this.surveyResults = results.surveyModules.filter(sm => sm.conceptId !== 1333342);
```
mhigbyflowers: do we want to keep this until we are done and ready for QA so it doesn't show up on the test??

mhigbyflowers (approved):
just one comment.
otherwise looks good

public-ui/src/app/data-browser/views/quick-search/quick-search.component.ts
SrushtiGangireddy: Thanks. Missed it as part of debugging.

PR #669, opened 2020-10-21
[risk=no] physical measurements data fix

SrushtiGangireddy:
Generate new data for prod so that current code does not break.


mhigbyflowers (approved):
LGTM

PR #670, opened 2020-10-22
[risk=no] Fix pm search

SrushtiGangireddy:
Fixing physical measurements search. 
Moving the concept fetch into pm component.


mhigbyflowers (approved):
looks cleaner, thanks!

PR #672, opened 2020-10-27
[RW-5772] [risk=no]  columns from the answers table COPE

mhigbyflowers:


SrushtiGangireddy:
This pr looks confusing with the rebase may be. Also the cope filter is being removed.

mhigbyflowers:
yeah idk what happen. Ill close and send another out.

PR #673, opened 2020-10-28
[RW-5601][risk=no] adding fitbit env var and removing workbench url

mhigbyflowers:
added fitbit env var 
also removed the workbench URL because it's not needed


SrushtiGangireddy (approved):
Except the flag flip looks good. thanks.

SrushtiGangireddy:
Should we add the flag to to be true in local env ?

PR #676, opened 2020-11-02
[RW-5815] changes to header

mhigbyflowers:
changes to header
see: https://precisionmedicineinitiative.atlassian.net/browse/RW-5815


SrushtiGangireddy (approved):
looks good to me. thanks.

PR #677, opened 2020-11-02
[risk=no] [RW-5771] removing %answered and participant count columns

SrushtiGangireddy:
Removing columns for cope survey.


mhigbyflowers (approved):
LGTM

PR #682, opened 2020-11-05
[risk=no] fix pm tile

SrushtiGangireddy:
Fix physical measurements tile.


mhigbyflowers:
image
this string is showing up without searching.
Other than that, 👍

mhigbyflowers (approved):
just one thing to fix

PR #684, opened 2020-11-06
[risk=no] fixing icon on labs and measurements

SrushtiGangireddy:
Adding data type icon on labs and measurements


mhigbyflowers (approved):
👍

PR #690, opened 2020-11-11
[risk=no] [RW-5802] Remove binning on question counts 

SrushtiGangireddy:
Cope survey version question counts are being binned. Remove binning.
Change family medical history text to family health history.


mhigbyflowers (approved):
👍  Looks great thanks!

PR #691, opened 2020-11-11
[risk=no] [RW-5883] Fixing few ui bugs

SrushtiGangireddy:
1. Fixing header look when disclaimer is popped up.
2. Fixing tooltip view.


SrushtiGangireddy:
Fixed this
image

image

Fixed this 
image

to 

image


to

mhigbyflowers (approved):
Looks good 👍 
these z-indexes are getting out of hand... we should look into them when we rewrite the UI

SrushtiGangireddy:
Yes we definitely need to. Thanks.

PR #694, opened 2020-11-17
[risk=no] cope version - use count value instead of source counts 

SrushtiGangireddy:
Use count value instead of source count.


mhigbyflowers (approved):
👍

PR #701, opened 2020-11-18
[risk=no] Survey question count fix

SrushtiGangireddy:
1. Removing question order number on surveys.
2. Counting all sub questions on question count.


mhigbyflowers (approved):
👍

PR #707, opened 2020-11-23
[risk=no] updating question name only for questions that are displayed on the app

SrushtiGangireddy:
Update question name to be similar to whatever is in survey pdf.


freemabd (approved):
Looks good.

PR #714, opened 2020-12-03
[risk=no] [RW-5988] adding % tooltip to COPE survey version answer graphs

SrushtiGangireddy:
image

Added '% of survey version count' to version answer chart.

### DIFF
--- a/public-ui/src/environments/environment.local.ts
+++ b/public-ui/src/environments/environment.local.ts
@@ -9,7 +9,7 @@ export const environment = {
   gtmId: 'GTM-NQ9XDTW',
   gtmAuth: 'kMy-ZUX8XsBfuanMZzP_5A',
   gtmPreview: 'env-25',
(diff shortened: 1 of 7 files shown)


public-ui/src/environments/environment.local.ts
```diff
@@ -9,7 +9,7 @@ export const environment = {
   gtmId: 'GTM-NQ9XDTW',
   gtmAuth: 'kMy-ZUX8XsBfuanMZzP_5A',
   gtmPreview: 'env-25',
-  copeFlag: false,
+  copeFlag: true,
```
SrushtiGangireddy, line 12: I forget to change this devel flag each time and find why cope is not showing up so leaving this to be true this time.
You can flip it back if you think leaving it on is a problem.

mhigbyflowers (approved):
👍 Thanks!

PR #716, opened 2020-12-03
[no-ticket][risk=no]fixing bug where fitbit chart tooltip is under biosex chart

mhigbyflowers:
fixing bug where Fitbit chart tooltip is under biosex chart


SrushtiGangireddy (approved):
LGTM. Thanks

PR #720, opened 2020-12-04
[risk=no] fixing post api error on search concepts api

SrushtiGangireddy:
Search concepts api has a error on content type. Fixing it by adding consumes type.


calbach (approved):
Note: if you ever add anymore `POST` endpoints you will need this as well. Looks like this is the only case currently.

PR #722, opened 2020-12-04
[risk=no]UI-improvements

mhigbyflowers:
- clicking the whole row in the survey that has subquestions now opens the subquestions along with the caret icon
- clicking question in surveys now opens the answer list
- tab focus now showing tooltip info (accessibility)
- tab accessibility added to Fitbit view
- loading spinner added to Fitbit view
- survey view has better responsiveness to window size
- Cope Survey Table is responsive to window size


SrushtiGangireddy (commented):
Looks good. Thanks.

SrushtiGangireddy (approved):

PR #727, opened 2020-12-07
[risk=no] flipping erwin release flags

SrushtiGangireddy:
Flipping cope and fitbit flags

### DIFF
--- a/public-ui/src/environments/environment.prod.ts
+++ b/public-ui/src/environments/environment.prod.ts
@@ -9,7 +9,6 @@ export const environment = {
   gtmAuth: 'ixtwsbMAbR3hx5st_ek2jg',
   gtmPreview: 'env-65',
   copeFlag: true,
-  preCopeFlag: true,
   fitbit: true,
   copeStacked: true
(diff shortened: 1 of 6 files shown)


public-ui/src/environments/environment.prod.ts
```diff
@@ -8,8 +8,8 @@ export const environment = {
   gtmId: 'GTM-NQ9XDTW',
   gtmAuth: 'ixtwsbMAbR3hx5st_ek2jg',
   gtmPreview: 'env-65',
-  copeFlag: false,
+  copeFlag: true,
   preCopeFlag: true,
```
mhigbyflowers: preCopeFlag should be false

mhigbyflowers (approved):
see the comment, otherwise 👍

public-ui/src/environments/environment.prod.ts
SrushtiGangireddy: I just saw that the precope flag is not used anywhere. Cleaned it up.

PR #741, opened 2020-12-17
[risk=no] [RW-5984] cleaning up code in common-api 

SrushtiGangireddy:
Moving code in common-api to public-api folder. 
cleaning up gradle files that refer to code in common-api.


freemabd (approved):
Seems fine.

PR #744, opened 2020-12-18
[risk=no] [RW-6078] API improvement - mapper for question concept

SrushtiGangireddy:
1. Creating mapper for question concept.
2. Removed TO_CLIENT_QUESTION_CONCEPT mapping.
3. Create question concept service.


public-api/src/main/java/org/pmiops/workbench/utils/mappers/QuestionConceptMapper.java
```diff
@@ -0,0 +1,34 @@
...
+import org.pmiops.workbench.model.QuestionConcept;
+import org.pmiops.workbench.utils.mappers.CommonMappers;
+import org.pmiops.workbench.utils.mappers.MapStructConfig;
+
+@Mapper(
+        config = MapStructConfig.class,
+        uses = {CommonMappers.class})
+public interface QuestionConceptMapper {
+    @Mapping(target = "conceptId", source = "db.dbQuestionConceptId.conceptId")
+    @Mapping(target = "conceptName", source = "db.conceptName")
```
freemabd: mappings are not required when properties have the same name.. only when they don't match. I would remove all line that fit that pattern.

freemabd (commented):
A comment about mapping names in the mapper and some failing test cases.

freemabd (approved):
Thanks

PR #745, opened 2020-12-28
API Improvement - Mapper for Achilles Dist Results

SrushtiGangireddy:
Moving Achilles Results Dist dao calls to service.


freemabd (approved):
Looks good.

PR #747, opened 2020-12-31
[risk=n0] cleaning up can show flag that filters surveys 

SrushtiGangireddy:
This flag was created when feature flags were not in place.
Cleaning it up as it is no longer needed.


freemabd (approved):
Looks good.

PR #749, opened 2021-01-05
[risk=no] Measurements filter fix

SrushtiGangireddy:
Reset the lab test/ order filter when the search term is changed.


mhigbyflowers (approved):
👍 LGTM

PR #751, opened 2021-01-06
[risk=no] [RW-6099] Remove terms from display on top terms lists

SrushtiGangireddy:
Filtering concepts Finding of abdominopelvic segment of trunk, Finding of substance level, General clinical state finding from display on tool.


mhigbyflowers (approved):
LGTM

PR #752, opened 2021-01-06
[risk=no] [RW-6099] fixing filter conditions

SrushtiGangireddy:
Removing previous filter concept row.
Adding fix for tooltip transparency ticket.
Removed inner scroll bar in sources chart.


mhigbyflowers (approved):
👍

PR #759, opened 2021-01-15
[risk=no] adding props file to codegen ignore

SrushtiGangireddy:
Adding properties file to swagger codegen ignore


mhigbyflowers (approved):
Looks good thanks

PR #763, opened 2021-01-22
[risk=no] [RW-6134] Fixing measurement filter bug 

SrushtiGangireddy:
Moving code from db-table component to ehr-view component.
Fixing measurement value filter.
It is easier this way and reduces the number of api calls for each filter change.


mhigbyflowers (approved):
👍 
please check these two links:

1. the top-concepts chart sill opens up a concept when clicked
2. Also the “explore source” button goes to a new page and it opens the concept

PR #770, opened 2021-02-01
[risk=no] Docker Sync Config

SrushtiGangireddy:
Setting up docker-sync in databrowser

Update docker-settings in preferences to use all CPUs and ~70-80% of available memory & swap

Install docker-sync
$ gem install --user-install docker-sync
Add ruby gems to PATH, instructions here https://docker-sync.readthedocs.io/en/latest/getting-started/installation.html

Checkout branch
$ git checkout master
$ git pull
$ git checkout docker_sync

$ cd public-api
$ ./project.rb docker-clean
Docker-sync will be started & stopped as part of the dev script
$ ./project.rb dev-up

Warm dev-up for realistic timings
$ ./project.rb dev-up

Note: Install docker-sync version 0.5.11 for similar configuration which requires specific version of libraries like docker-compose and ruby (~>2.0).

Check here for any dependencies. https://rubygems.org/gems/docker-sync/versions/0.5.11

### DIFF
--- a/public-api/docker-compose.yaml
+++ b/public-api/docker-compose.yaml
@@ -11,7 +11,7 @@ x-api-defaults: &api-defaults
   # it from the provided build context or use a cached local version.
   # When making changes to this image, you can modify this tag to force all devs
   # to rebuild.
-  image: allofustest/workbench-dev-api:local-3
+  image: allofustest/workbench-dev-api:local-2
   build:
     context: ./src/dev/server
   user: ${UID}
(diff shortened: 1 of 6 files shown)


public-api/docker-compose.yaml
```diff
@@ -11,7 +11,7 @@ x-api-defaults: &api-defaults
   # it from the provided build context or use a cached local version.
   # When making changes to this image, you can modify this tag to force all devs
   # to rebuild.
-  image: allofustest/workbench-dev-api:local-2
+  image: allofustest/workbench-dev-api:local-3
```
calbach: nit: you shouldn't need to bump this unless you changed the Dockerfile

calbach (approved):

PR #772, opened 2021-02-04
[risk=no] [RW-6186] Change the default unit order of height pm

SrushtiGangireddy:
Make centimeter fist default unit in height physical measurement.


mhigbyflowers (approved):
👍

PR #775, opened 2021-02-10
[risk=no] [RW-6217] React Conversion - Statement popup Modal + button

SrushtiGangireddy:
Creating react pop up modal component and button component.
Switching between angular and react components using env flag.


public-ui/src/app/shared/components/button/ReactButtonComponent.tsx
```diff
@@ -0,0 +1,57 @@
...
+  ViewEncapsulation
+} from '@angular/core';
+import * as React from 'react';
+import { FunctionComponent } from 'react';
+import * as ReactDOM from 'react-dom';
+import { BaseReactWrapper } from '../../../data-browser/base-react/base-react.wrapper';
+
+const containerElementName = 'myReactComponentContainer';
+
+const ReactButtonComponent =
```
dolbeew: I'm not sure this component is necessary. Seems like it renders a standard HTML button, which can be used as is in both Angular and React.

public-ui/src/app/shared/components/pop-up/PopUpReactComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+    };
+
+@Component({
+  selector: 'app-popup-react',
+  template: `<span #${containerElementName}></span>`,
+  styleUrls: ['../../../styles/template.css', './pop-up.component.css'],
+  encapsulation: ViewEncapsulation.None,
+})
+
+export class PopUpWrapperComponent extends BaseReactWrapper {
```
dolbeew, line 78: The Angular wrapper for each component should basically be empty except for a list of Inputs that will be the props for the React component. I think most of this can either be moved into the `BaseReactWrapper` class or moved into the React component itself.

public-ui/src/app/shared/components/pop-up/PopUpReactComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+  @ViewChild(containerElementName, { static: true }) containerRef: ElementRef;
+  @Input() public title = 'test title';
+  @Input() public statement;
+  @Output() public close = new EventEmitter<void>();
```
dolbeew: When converting `@Output`s to React, we need to switch to an `@Input` that passes a callback function. It should do the same thing and eliminate the need for the `closeClick` function below.

public-ui/src/app/shared/components/pop-up/PopUpReactComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { BaseReactWrapper } from '../../../data-browser/base-react/base-react.wrapper';
+const containerElementName = 'myReactComponentContainer';
+
+const PopUpReactComponent =
+    (props) => {
+    return <div className='data-statement'>
+            <div className='card'>
+                <div onClick={props.closeClick} className='close'>x</div>
+                   <h2 className='card-title'>{props.title}</h2>
+                   <div className='card-body' dangerouslySetInnerHTML=
```
dolbeew: Using `dangerouslySetInnerHTML` should be avoided if possible. Might be better to convert the `HelpTextService` to a constant since it seems to be static content and call it directly here.

public-ui/src/app/shared/components/pop-up/PopUpReactComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public closeClick() {
+      if (this.close) {
+        this.close.emit();
+        this.render();
+      }
+    }
+
+  public render() {
```
dolbeew: With some changes, this should be able to move into the `BaseReactWrapper`

dolbeew (commented):
I left a few comments that may need more explanation. If you and @mhigbyflowers have some time early next week, maybe we can do a call to discuss further. I'm going to see if I can make some of the changes locally just to make sure they will work.

public-ui/src/app/shared/components/pop-up/PopUpReactComponent.tsx
SrushtiGangireddy: Moved this. Thanks.

public-ui/src/app/shared/components/pop-up/PopUpReactComponent.tsx
SrushtiGangireddy: Added helptext constants in here.

public-ui/src/app/shared/components/pop-up/PopUpReactComponent.tsx
SrushtiGangireddy: Thanks. Addressed this.

public-ui/src/app/shared/components/pop-up/PopUpReactComponent.tsx
SrushtiGangireddy: Sounds good. Thanks.

public-ui/src/app/shared/components/button/ReactButtonComponent.tsx
SrushtiGangireddy: Made the change.

SrushtiGangireddy:
@dolbeew Made few changes. Moving render to the base component seems to be little complicated. I need your help declaring the wrapper component and injecting it from the child components.

PR #776, opened 2021-02-10
[risk=no] [RW-6150] Removing common-ui

SrushtiGangireddy:
Removing code from common-ui.
Moved it to public-ui.


mhigbyflowers (approved):
looks good 👍

PR #778, opened 2021-02-12
[RW-6223][risk=no] react ResultLinksComponent

mhigbyflowers:


mhigbyflowers:
There are more components like Tool-tips that need to be plugged in when they are finished. This is just a subbed out component and will add functionality when those components are available.
Also need to create a router so it can link out.

SrushtiGangireddy (approved):
Looks good to me by itself. Does this has to be linked with router later ?

PR #784, opened 2021-02-18
[risk=no] [DB-1014] fixing cope visit from url

SrushtiGangireddy:
Cope visit from url keeps spinning forever.


public-ui/src/app/utils/db-config.service.ts
```diff
@@ -157,7 +157,7 @@ export class DbConfigService {
     { conceptId: 43529712, conceptName: 'Personal Medical History' },
     { conceptId: 43528895, conceptName: 'Health Care Access and Utilization' },
     { conceptId: 43528698, conceptName: 'Family Health History' },
-    { conceptId: 1333342, conceptName: 'COVID-19 Participant Experience' }
+    { conceptId: 1333342, conceptName: 'COVID-19 Participant Experience (COPE)' }
```
SrushtiGangireddy, line 160: This alone should be fixing the visit. But I want to check and make sure things are working.

mhigbyflowers (approved):
Looks good, thanks

PR #786, opened 2021-02-22
[RW-6222][risk=no] Convert footer component to React

dolbeew:
- Convert `rh-footer.component` to React
- Move footer content from `index.html` to new footer component


public-ui/src/app/shared/services/header-footer.service.ts
```diff
@@ -1,6 +1,119 @@
 import { Injectable } from '@angular/core';
 import { environment } from '../../../environments/environment';
 
+export const menuItems = [
```
dolbeew, line 4: When copying this into a const, I removed several objects and properties that were unused in the footer. We can add them back later if needed for other components.

Also consolidated `sub0`, `sub1`, etc. into `submenu` and just check the length to see if it has a submenu. The React template didn't like iterating over objects with different properties (`sub0` vs `sub1`)

public-ui/src/app/views/app/app.component.html
```diff
@@ -10,4 +10,4 @@
 		</div> -->
   <!-- .cta-content -->
 </main>
-<app-rh-footer></app-rh-footer>
\ No newline at end of file
+<app-rh-footer-react></app-rh-footer-react>
```
dolbeew, line 13: Just replaced with the React component here since I don't think we can conditionally show/hide the footer in `index.html`

public-ui/src/app/shared/components/rh-footer/rh-footer.tsx
```diff
@@ -0,0 +1,129 @@
...
+              Learn more about the program protocol, leadership, and governance at&nbsp;
+              <a href='https://allofus.nih.gov'
+                 target='_blank'
+                 rel='noopener'>AllofUs.nih.gov</a>
+            </p>
+          </div>
+          <div className='cta-col'>
+            <p>Follow <i>All of Us</i></p>
+            <ul id='menu-social' className='social-links-menu'>
+              {socialLinks.map((item, i) => <li key={i}>
```
dolbeew, line 79: Moved the strings for the social icons into a const so we can just iterate here

SrushtiGangireddy (approved):
Thanks. Looks good to me.

PR #788, opened 2021-02-23
[risk=no] [RW-5994] Highlight search react conversion

SrushtiGangireddy:
Converting highlight search component.


dolbeew (approved):
LGTM

I think once we start converting the parent components (`ehr-view`, `fmh-view`, `survey-view`), we could probably just make `highlightSearchTerm` a helper function that gets called directly in the render function of those components. No need to have a component that just returns the output of `highlightSearchTerm` at that point.

SrushtiGangireddy:
Thanks Will. That was what i was thinking too.

PR #792, opened 2021-02-26
[RW-6219][risk=no] Convert header component to React

dolbeew:
Convert `rh-header` component to React

### DIFF
--- a/public-ui/src/app/shared/components/rh-header/rh-header.component.css
+++ b/public-ui/src/app/shared/components/rh-header/rh-header.component.css
@@ -562,7 +562,11 @@ li.active-menu:before {
     display: none;
   }
   .site-nav-container nav.main-navigation .menu-primary-container ul.main-menu {
-    display: none;
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+    margin-block-end: 0;
+    width: 100%;
   }
   .site-nav-container nav.main-navigation .menu-primary-container ul.main-menu li.icon_search:hover ul.sub-menu,
   .site-nav-container nav.main-navigation .menu-primary-container ul.main-menu li.icon_search:focus + ul.sub-menu {
@@ -601,18 +605,8 @@ li.active-menu:before {
   .cta-bar aside p {
     white-space: nowrap;
   }
-  .mobile-nav {
-    display: none;
-  }
   .cta-bar .login-btn {
     display: block;
-  }
-  .site-nav-container nav.main-navigation .menu-primary-container ul.main-menu {
-    display: flex;
-    flex-direction: row;
-    justify-content: flex-end;
-    margin-block-end: 0;
-    width: 100%;
   }
   .site-nav-container nav.main-navigation .menu-primary-container ul.main-menu li {
     border: none;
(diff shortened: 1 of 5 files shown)


public-ui/src/app/shared/components/rh-header/rh-header.component.css
```diff
@@ -562,7 +562,11 @@ li.active-menu:before {
     display: none;
   }
   .site-nav-container nav.main-navigation .menu-primary-container ul.main-menu {
-    display: none;
+    display: flex;
```
dolbeew, line 565: When testing the display of the mobile menu, I found if you increase the width of the screen with the mobile menu open, there's a middle ground where both menus are displayed:
Screen Shot 2021-02-25 at 11 46 35 PM
So I made a few changes that _should_ make it display only one or the other.

public-ui/src/app/shared/components/rh-header/rh-header.tsx
```diff
@@ -0,0 +1,163 @@
...
+          </div>
+        </div>
+        {menuOpen && <ul id='menu-mobile'>
+          {menuItems.map((menu, m) =>
+            <li className={`main-menu-item menu-item-has-children ${menu.title} ${submenusOpen.includes(menu.title) ? 'flip' : ''}`}
+                onClick={() => this.toggleMobileSubmenu(menu.title)}>
+              <a href={menu.url} style={{color: '#2b266d'}}>{menu.title}</a>
+              {submenusOpen.includes(menu.title) && menu.submenu.length > 0 && <ul className='sub-menu'>
+                {menu.submenu.map((sub, s) => <li key={s} className='menu-item'>
+                  <a href={sub.url} style={{color: '#2b266d'}}>{sub.title}</a>
```
dolbeew, line 124: I was seeing some inconsistencies with the text color of the links, seems sometimes it was reading styles from the footer css file for some reason. I set an inline style on both links here to make sure the correct color is shown.

dolbeew (commented):
I tested out the [Primereact menubar](https://primefaces.org/primereact/showcase/#/menubar) initially but the submenus would only open on click, not on hover like the current menu, so I ended up converting the current template.

SrushtiGangireddy (approved):
Looks good to me. Thanks.

mhigbyflowers (commented):
image

- A few of the links in the mobile menu are a light blue
- Also when the links are hovered over they turn white.

dolbeew:
@mhigbyflowers I believe those are existing issues that should be fixed in the React version. Do you have `testReact: true` set locally? I left it set to `false` since that's what it's currently set to in master.

mhigbyflowers:
@dolbeew  Ah yeah I forgot to change the env var. I was checking it against prod, where those issues are not present.

public-ui/src/app/shared/components/rh-header/rh-header.tsx
```diff
@@ -0,0 +1,165 @@
...
+    this.setState({submenusOpen: updatedSubmenusOpen});
+  }
+
+  render() {
+    const {menuOpen, searchTerms, submenusOpen} = this.state;
+    return <React.Fragment>
+      <div className='cta-bar'>
+        <a className='login-btn'
+           href='https://workbench.researchallofus.org/login'
+           style={{width: 'auto'}}>
```
dolbeew, line 38: Had to use an inline style here also since it was applying `width: 100%` from the footer CSS

PR #806, opened 2021-03-16
[risk=no] Fixing survey sub question bug

SrushtiGangireddy:
Fixing survey sub question bug


mhigbyflowers (approved):
Thanks for finding this so quick 👍

PR #814, opened 2021-03-25
[risk=no] [RW-6513] error message react conversion

SrushtiGangireddy:
Converting error-message component to react


public-ui/src/app/data-browser/views/error-message/error-message-react.component.tsx
```diff
@@ -0,0 +1,55 @@
...
+export class ErrorMessageReactComponent extends React.Component<Props, {}> {
+
+  constructor(props: Props) {
+    super(props);
+  }
+
+  render() {
+    const {dataType} = this.props;
+    const iconClass = 'alert-icon';
+    const iconShape = 'exclamation-triangle';
```
dolbeew: Nit: Since these are only used once, just put them directly on the element.

public-ui/src/app/data-browser/views/error-message/error-message-react.component.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const alertText =  (dataType === 'data') ? '<span className="alert-text"></span>' : '<span className="alert-text">' +
+                                'Sorry, the chart cannot be displayed. Please try refreshing the page.' +
+                              '</span>';
```
dolbeew: This is currently unused. We should either call this in the template or remove it.

public-ui/src/app/data-browser/views/error-message/error-message-react.component.tsx
```diff
@@ -0,0 +1,55 @@
...
+    return <div className='alert alert-box'>
+      <div className='alert-items'>
+        <div className='alert-item static'>
+          <div className='alert-icon-wrapper'>
+            <ClrIcon shape={iconShape} className={iconClass}
+                            style={{width: 26, height: 26}} />
+          </div>
+          <span className='alert-text'>
+          {dataType === 'data' ? 'Sorry, the data are currently unavailable. Please try refreshing the page or returning home <a [routerLink]="[\'\']">here.</a>'
```
dolbeew: I realize that currently, the `dataType` will never be 'data' for this component, but we shouldn't leave `[routerLink]="[\'\']"` here.

public-ui/src/app/data-browser/views/error-message/error-message-react.component.tsx
SrushtiGangireddy: Changed this.

public-ui/src/app/data-browser/views/error-message/error-message-react.component.tsx
SrushtiGangireddy: Sorry this should not be in here.

public-ui/src/app/data-browser/views/error-message/error-message-react.component.tsx
SrushtiGangireddy: So, the datatype would be data from here https://github.com/all-of-us/data-browser/blob/master/public-ui/src/app/data-browser/views/ehr-view/ehr-view.component.html#L32. 

I am assuming we would have to wait till we figure out react router to place home link in here.

public-ui/src/app/data-browser/views/error-message/error-message-react.component.tsx
dolbeew: Can we remove these since they're now unused?

dolbeew (approved):

PR #820, opened 2021-04-02
[risk=no] Survey desc comp

SrushtiGangireddy:
Pulling out simple survey desc into different component.
This can be part of any parent component that has smaller survey-view parts in future.


mhigbyflowers (approved):
lgtm Thanks

PR #823, opened 2021-04-05
[risk=no] Survey chart component conversion wip

SrushtiGangireddy:
Converting survey chart component

### DIFF
--- a/public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
+++ b/public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
@@ -4,11 +4,11 @@ import {
   ViewEncapsulation
 } from '@angular/core';
 import { BaseReactWrapper } from 'app/data-browser/base-react/base-react.wrapper';
+import { AgeChartReactComponent } from 'app/data-browser/charts/chart-age/chart-age-react.component';
+import { BioSexChartReactComponent } from 'app/data-browser/charts/chart-biosex/chart-biosex-react.component';
+import { VersionChartReactComponent } from 'app/data-browser/charts/chart-version/chart-version-react.component';
 import { TooltipReactComponent } from 'app/data-browser/components/tooltip/tooltip-react.component';
 import { ErrorMessageReactComponent } from 'app/data-browser/views/error-message/error-message-react.component';
-import { BioSexChartReactComponent } from 'app/data-browser/charts/chart-biosex/chart-biosex-react.component';
-import { AgeChartReactComponent } from 'app/data-browser/charts/chart-age/chart-age-react.component';
-import { VersionChartReactComponent } from 'app/data-browser/charts/chart-version/chart-version-react.component';
 import { GraphType } from 'app/utils/enum-defs';
 import { triggerEvent } from 'app/utils/google_analytics';
 import * as React from 'react';
@@ -106,26 +106,26 @@ export class SurveyChartReactComponent extends React.Component<Props, State> {
             }
         </div>
 {this.state.displayGraphErrorMessage
-    ? <div className="graph-error-message">
+    ? <div className='graph-error-message'>
                   <ErrorMessageReactComponent dataType='chart'></ErrorMessageReactComponent>
                 </div>
     : [
         isLoaded && this.state.selectedChartAnalysis.analysisId === 3111 ?
-            <div className="chart" key='biosex-chart'>
+            <div className='chart' key='biosex-chart'>
              <BioSexChartReactComponent
              domain='survey' genderAnalysis={this.state.selectedChartAnalysis}
              genderCountAnalysis={this.props.surveyCountAnalysis.genderCountAnalysis}
              selectedResult={this.props.selectedResult}></BioSexChartReactComponent>
             </div> : [
             isLoaded && this.state.selectedChartAnalysis.analysisId === 3112 ?
-            <div className="chart" key='age-chart'>
+            <div className='chart' key='age-chart'>
                                  <AgeChartReactComponent
                                  domain='survey' ageAnalysis={this.state.selectedChartAnalysis}
                                  ageCountAnalysis={this.props.surveyCountAnalysis.ageCountAnalysis}
                                  selectedResult={this.props.selectedResult}></AgeChartReactComponent>
             </div> : [
             isLoaded && this.state.selectedChartAnalysis.analysisId === 3113 ?
-            <div className="chart" key='age-chart'>
+            <div className='chart' key='age-chart'>
             <VersionChartReactComponent versionAnalysis={this.state.selectedChartAnalysis}
             surveyVersionAnalysis={this.props.versionAnalysis}
             selectedResult={this.props.selectedResult}></VersionChartReactComponent>
--- a/public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
+++ b/public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
@@ -8,6 +8,7 @@ import { TooltipReactComponent } from 'app/data-browser/components/tooltip/toolt
 import { ErrorMessageReactComponent } from 'app/data-browser/views/error-message/error-message-react.component';
 import { BioSexChartReactComponent } from 'app/data-browser/charts/chart-biosex/chart-biosex-react.component';
 import { AgeChartReactComponent } from 'app/data-browser/charts/chart-age/chart-age-react.component';
+import { VersionChartReactComponent } from 'app/data-browser/charts/chart-version/chart-version-react.component';
 import { GraphType } from 'app/utils/enum-defs';
 import { triggerEvent } from 'app/utils/google_analytics';
 import * as React from 'react';
@@ -36,7 +37,8 @@ interface Props {
 export class SurveyChartReactComponent extends React.Component<Props, State> {
   constructor(props) {
     super(props);
-    this.state = {graphToShow: this.props.isCopeSurvey? GraphType.SurveyVersion: GraphType.BiologicalSex, displayGraphErrorMessage: false, isLoaded: false, selectedChartAnalysis: null};
+    this.state = {graphToShow: this.props.isCopeSurvey? GraphType.SurveyVersion: GraphType.BiologicalSex,
+    displayGraphErrorMessage: false, isLoaded: false, selectedChartAnalysis: null};
   }
 
   componentDidMount() {
@@ -45,12 +47,8 @@ export class SurveyChartReactComponent extends React.Component<Props, State> {
 
   selectGraphType(g: any, q: any, answer: any) {
     const {surveyName, searchTerm} = this.props;
-    const {graphToShow} = this.state;
-    q.graphToShow = graphToShow;
-    if (q.graphDataToShow === null) {
-        q.graphDataToShow = 'Count';
-    }
-    switch (graphToShow) {
+    q.graphToShow = g;
+    switch (g) {
         case GraphType.AgeWhenSurveyWasTaken:
             q.selectedAnalysis = q.ageAnalysis;
             break;
@@ -81,7 +79,8 @@ export class SurveyChartReactComponent extends React.Component<Props, State> {
 
   getLabel(g) {
     const {surveyName, question, answer} = this.props;
-    return surveyName + ' - Q' +  question.actualQuestionNumber+ ' - ' +  question.conceptName + ' - ' + answer.stratum4 + ' - ' + g;
+    return surveyName + ' - Q' +  question.actualQuestionNumber+ ' - ' +
+    question.conceptName + ' - ' + answer.stratum4 + ' - ' + g;
   }
 
 
@@ -99,7 +98,8 @@ export class SurveyChartReactComponent extends React.Component<Props, State> {
                  tabIndex={tabIndex} key={index}>
                  <span>{g}</span>
                  <TooltipReactComponent tooltipKey={this.getTooltipKey(g)}
-                 label={this.getLabel(g)} searchTerm={searchTerm} action='Survey Chart Tooltip'></TooltipReactComponent>
+                 label={this.getLabel(g)} searchTerm={searchTerm} action='Survey Chart Tooltip'>
+                 </TooltipReactComponent>
                  </div>
                  );
               })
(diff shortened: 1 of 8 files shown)


public-ui/src/app/data-browser/views/survey-view/survey-view.component.html
```diff
@@ -201,15 +201,17 @@ <h2 class="secondary-display">
...
                       #chartElement>
-                      <app-survey-chart [isCopeSurvey]="isCopeSurvey" [graphButtons]="graphButtons" [question]="q"
-                        [selectedAnalysis]="q.selectedAnalysis" [selectedResult]="a" [surveyName]="survey.name" [versionAnalysis]='surveyVersions'
-                        [surveyCountAnalysis]="surveyCountAnalysis" [searchTerm]="prevSearchText" [answer]="a">
-                      </app-survey-chart>
+                      <app-survey-chart-react [graphButtons]="graphButtons" [isCopeSurvey]="isCopeSurvey" [question]="q" [answer]="a"
+                                              [selectedAnalysis]="q.selectedAnalysis" [selectedResult]="a"
+                                              [surveyName]="survey.name" [versionAnalysis]='surveyVersions'
+                                              [surveyCountAnalysis]="surveyCountAnalysis" [searchTerm]="prevSearchText"
+                                              (click)="$event.stopPropagation();"></app-survey-chart-react>
```
SrushtiGangireddy: Had to have this in here to stop closing the survey answer tab.

dolbeew: I may have found a temp workaround so we don't need this. Try adding an optional event param to `showAnswerGraphs` at line 520 in `survey-view.component.ts`, then we can disregard clicks from within the expanded area:
```
public showAnswerGraphs(a: any, q: any, level?: number, event?: MouseEvent) {
    if (event && this.chartEl && this.chartEl.nativeElement.contains(event.target)) {
      return;
    }
```

Then we just need to pass the click event at line 140 in this file:
```
<div [ngClass]="a.expanded?'survey-tbl-r-state':''" class="survey-tbl-exp-r "
                    *ngFor="let a of q.countAnalysis.results" (click)="showAnswerGraphs(a,q,1, $event)"
                    [hidden]="!showAnswer[q.conceptId]">
```
Seems to be working for me, if you could test it to verify it works as intended.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,179 @@
...
+    selectedChartAnalysis: any;
+    displayGraphErrorMessage: boolean;
+    isLoaded: boolean;
+}
+
+interface Props {
+    graphButtons: any;
+    question: any;
+    answer: any;
+    selectedAnalysis: any;
```
dolbeew: Can remove this, doesn't seem to be referenced anywhere

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,179 @@
...
+    surveyCountAnalysis: any;
+    isCopeSurvey: any;
+    versionAnalysis: any;
+}
+
+export class SurveyChartReactComponent extends React.Component<Props, State> {
+  constructor(props) {
+    super(props);
+    this.state = {graphToShow: this.props.isCopeSurvey ? GraphType.SurveyVersion : GraphType.BiologicalSex,
+    displayGraphErrorMessage: false, isLoaded: false, selectedChartAnalysis: null};
```
dolbeew: Nit: for a multi-line object, put each property on its own line:
```suggestion
    this.state = {
      graphToShow: this.props.isCopeSurvey ? GraphType.SurveyVersion : GraphType.BiologicalSex,
      displayGraphErrorMessage: false,
      isLoaded: false,
      selectedChartAnalysis: null
    };
```

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,179 @@
...
+                 label={this.getLabel(g)} searchTerm={searchTerm} action='Survey Chart Tooltip'>
+                 </TooltipReactComponent>
+                 </div>
+                 );
+              })
+            }
+        </div>
+{displayGraphErrorMessage
+    ? <div className='graph-error-message'>
+                  <ErrorMessageReactComponent dataType='chart'></ErrorMessageReactComponent>
```
dolbeew: Use self-closing syntax if there's no nested content:
```suggestion
                  <ErrorMessageReactComponent dataType='chart'/>
```

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,179 @@
...
+                </div>
+    : [
+        isLoaded && selectedChartAnalysis.analysisId === 3111 ?
+            <div className='chart' key='biosex-chart'>
+             <BioSexChartReactComponent
+             domain='survey' genderAnalysis={selectedChartAnalysis}
+             genderCountAnalysis={surveyCountAnalysis.genderCountAnalysis}
+             selectedResult={selectedResult}></BioSexChartReactComponent>
```
dolbeew: Use self-closing syntax

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,179 @@
...
+            </div> : [
+            isLoaded && selectedChartAnalysis.analysisId === 3112 ?
+            <div className='chart' key='age-chart'>
+                                 <AgeChartReactComponent
+                                 domain='survey' ageAnalysis={selectedChartAnalysis}
+                                 ageCountAnalysis={surveyCountAnalysis.ageCountAnalysis}
+                                 selectedResult={selectedResult}></AgeChartReactComponent>
```
dolbeew: Use self-closing syntax

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,179 @@
...
+            </div> : [
+            isLoaded && selectedChartAnalysis.analysisId === 3113 ?
+            <div className='chart' key='age-chart'>
+            <VersionChartReactComponent versionAnalysis={selectedChartAnalysis}
+            surveyVersionAnalysis={versionAnalysis}
+            selectedResult={selectedResult}></VersionChartReactComponent>
```
dolbeew: Use self-closing syntax

public-ui/src/app/data-browser/views/survey-view/survey-view.component.ts
```diff
@@ -517,7 +517,19 @@ export class SurveyViewComponent implements OnInit, OnDestroy {
       });
   }
 
-  public showAnswerGraphs(a: any, q: any, level?: number) {
+  public showAnswerGraphs(a: any, q: any, level?: number, event?: MouseEvent) {
+    if (event && this.chartEl && this.chartEl.nativeElement.contains(event.target)) {
```
dolbeew: This only works for one section per question since each element in the loop gets the same selector (sorry, didn't notice this when I first tested this workaround). If you expand a second answer under the same question, it will close when you try to click the graph tabs. It should work if we use `@ViewChildren` decorator instead.

On line 58:
`@ViewChildren('chartElement') chartEl: QueryList<ElementRef>;`

Then here, we need to check if any element in the list is clicked:
```suggestion
    if (event && this.chartEl && this.chartEl.some(el => el.nativeElement.contains(event.target))) {
```

This seems to work well, but after changing, I noticed if multiple answers are expanded, switching graph types in one answer will switch for all answers that are expanded under the same question:

https://user-images.githubusercontent.com/40036095/113972560-69f4b980-9800-11eb-8912-66dd3cc29643.mov

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,179 @@
...
+                 </div>
+                 );
+              })
+            }
+        </div>
+{displayGraphErrorMessage
+    ? <div className='graph-error-message'>
+                  <ErrorMessageReactComponent dataType='chart'></ErrorMessageReactComponent>
+                </div>
+    : [
```
dolbeew: Can also remove all `[]` brackets from template, should work fine without them

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: removed thanks.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: changed thanks.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: changed thanks.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: changed thanks.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: changed thanks.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: changed thanks.

public-ui/src/app/data-browser/views/survey-view/survey-view.component.ts
SrushtiGangireddy: Thanks for this fix. 
It also looks like all the charts have same analysis when multiple charts are opened. So seems like the problems with sharing state across charts.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+            q.selectedAnalysis = q.genderAnalysis;
+            break;
+    }
+    this.setState({
+        graphToShow: g,
+        selectedChartAnalysis: q.selectedAnalysis,
+        isLoaded: true,
+        displayGraphErrorMessage: q.selectedAnalysis === undefined ||
+        (q.selectedAnalysis && q.selectedAnalysis.results.filter(a => a.stratum3 ===
+        answer.stratum3).length === 0) });
```
dolbeew: I think modifying `q` here is causing the other answer graphs to change. Since `q` is the `question` prop and is passed to this function directly in `render`, any changes to it will get passed up to the parent through props, then down to the other charts.

I made these changes and seemed to at least be showing different data on each chart:
```suggestion
    let selectedAnalysis;
    switch (g) {
        case GraphType.AgeWhenSurveyWasTaken:
            selectedAnalysis = q.ageAnalysis;
            break;
        case GraphType.SurveyVersion:
            selectedAnalysis = q.versionAnalysis;
            break;
        default:
            selectedAnalysis = q.genderAnalysis;
            break;
    }
    this.setState({
        graphToShow: g,
        selectedChartAnalysis: selectedAnalysis,
        isLoaded: true,
        displayGraphErrorMessage: selectedAnalysis === undefined ||
        (selectedAnalysis && selectedAnalysis.results.filter(a => a.stratum3 ===
        answer.stratum3).length === 0) });
```

SrushtiGangireddy: Thanks, it also does not seem to flipping the charts with other chart button click.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+        isLoaded: false,
+        selectedChartAnalysis: null,
+    };
+  }
+
+  componentDidMount() {
+    this.selectGraphType(this.state.graphToShow, this.props.question, this.props.answer);
+  }
+
+  selectGraphType(g: any, q: any, answer: any) {
```
dolbeew: Nit: since `q` and `answer` are just passed from props, would be simpler to access them from props here instead of passing as args. Then just need to pass a single arg.

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+      triggerEvent('conceptClick', 'More than one race /ethncitiy graph view',
+        'Expand to see graphs', surveyName + ' - Q'
+        + q.actualQuestionNumber + ' - ' + q.conceptName + ' - ' + answer.stratum4 +
+        ' - ' + g, searchTerm, null);
+    }
+    triggerEvent('conceptClick', 'View Graphs',
+      'Expand to see graphs', surveyName + ' - Q'
+      + q.actualQuestionNumber + ' - ' + q.conceptName + ' - ' + answer.stratum4 +
+      ' - ' + g, searchTerm, null);
+    q.graphToShow = g;
```
dolbeew: Do we need this? Seems like we could just use `g` below (line 83) instead of `q.graphToShow`

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+  selector: 'app-survey-chart-react',
+  template: `<span #root></span>`,
+  styleUrls: ['./survey-chart.component.css', '../../../styles/template.css', '../../../styles/page.css'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class SurveyChartWrapperComponent extends BaseReactWrapper {
+  @Input() graphButtons: string[];
+  @Input() question: any;
+  @Input() answer: any;
+  @Input() selectedAnalysis: any;
```
dolbeew: Do we need this? Doesn't seem to be used in the component

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: changed this thanks

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: changed this thanks

public-ui/src/app/data-browser/views/survey-chart/survey-chart-react.component.tsx
SrushtiGangireddy: changed this thanks

dolbeew (approved):

PR #827, opened 2021-04-07
[risk=no] [RW-6563] replacing spinner component with react one

SrushtiGangireddy:
Converting spinner component to react


dolbeew (commented):
We should put simple components like this in utility file, similar to [spinners.tsx](https://github.com/all-of-us/workbench/blob/master/ui/src/app/components/spinners.tsx) in workbench. Obviously we won't be able to use them yet, but we can go ahead and create them, then switch when the parent components are converted.

We'd also need to use inline or object-based styles instead of CSS files, which we probably need to do anyway. Currently, we're still relying on Angular to load the CSS files for wrapped React components.

public-ui/src/app/utils/spinner.tsx
```diff
@@ -0,0 +1,70 @@
...
+    from{background:transparent}
+    to{background:#302c70}
+  }
+`;
+
+
+interface SpinnerProps {
+  loading: boolean;
+  dots: boolean;
+}
```
dolbeew: I think we can remove both props. `loading` should be handled in the parent and only render the spinner if true. Also, we should split the two types into separate consts. 

So once we convert the parent components, should look something like this:
```
{loading && <Spinner/>}
{loading && <LoadingDots/>} // or whatever we choose to name this one
```

public-ui/src/app/utils/spinner.tsx
```diff
@@ -0,0 +1,70 @@
+import * as React from 'react';
+
+export const spinnerCss = `
```
dolbeew, line 12: This is fine for now, but may change when we start aligning styles more with workbench. May be good to separate styles used for the different spinners if we split them into separate consts

public-ui/src/app/utils/spinner.tsx
SrushtiGangireddy: I split these both into different consts.

public-ui/src/app/utils/spinner.tsx
SrushtiGangireddy: I have seperated styles. Ll take a look at workbench styles.

public-ui/src/app/utils/spinner.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <div className='loading-dots'>
+            <div className='dot'></div>
+                <div className='dot'></div>
+                <div className='dot'></div>
+                <div className='dot'></div>
+           </div>
+           </React.Fragment>;
+};
+
+export const SpinnerContainer = ({loading, dots}) => {
```
dolbeew: So, I think we can remove this also. The components above look good and we can just import them individually whenever we need them in other components.

public-ui/src/app/utils/spinner.tsx
dolbeew: Yeah, I'm going to try to work on styles some next week. Converting the dots CSS here to objects looks pretty straightforward, not sure about some of the spinner styles like `@keyframes` and `a:link,a:visited`.

SrushtiGangireddy:
yeah for keyframes i see there are some options like using styles-components or css-modules. We can try them.

public-ui/src/app/utils/spinner.tsx
SrushtiGangireddy: Got it. thanks.

dolbeew (approved):

PR #840, opened 2021-04-29
[RW-6612][risk=no] Add React style utils, convert rh-footer styles

dolbeew:
- Adds `reactStyles` helper to build styles with JS objects
- Converts styles for `rh-footer` component using a combination of JS objects and CSS strings
- Drops the style sheet import from Angular wrapper

Per our meeting a couple weeks ago, we should try to use this method of styles for new conversions - JS objects whenever possible and CSS string in a `<style></style>` for media queries, pseudo classes, etc.


public-ui/src/app/shared/components/rh-footer/rh-footer.tsx
```diff
@@ -1,27 +1,155 @@
...
+    font-size: 0.7em;
+    color: white;
+    line-height: 1;
+    width: 100%;
+  }
+  .login-btn a:hover {
+    background-color: rgba(190, 225, 255, 1);
+    color: #262262;
+  }
+  .logo {
```
SrushtiGangireddy: Can this be removed ? We seem to be using style from reactstyles for this class.

SrushtiGangireddy (approved):
LGTM. Thanks.

public-ui/src/app/shared/components/rh-footer/rh-footer.tsx
dolbeew: Nice catch. I ended up removing the JS style since this one changes based on screen size. Also renamed the class to `footer-logo` since it was getting styles applied to it from the `logo` class from `rh-header` component.

mhigbyflowers (approved):
LGTM Thanks for adding this.

PR #842, opened 2021-04-29
[RW-6264][risk=no] Survey-question 

mhigbyflowers:
image


public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,138 @@
...
+                    this.props.question.countAnalysis.results.sort((a1, a2) => {
+                        if (a1.countValue > a2.countValue) {
+                            return -1;
+                        }
+                        if (a1.countValue < a2.countValue) {
+                            return 1;
+                        }
+                        return 0;
+                    });
+                    this.setState({ fetchComplete: true });
```
dolbeew: This will usually work, but modifying a components own props is [strongly discouraged](https://reactjs.org/docs/components-and-props.html#props-are-read-only) by React. A better way would be to create a variable in state to store the updated question and pass that to `SurveyAnswerReactComponent`. This would also allow us to drop the `fetchComplete` state var and just check if the updated question is set in state:
```suggestion
                  const questionWithResults = {
                        countAnalysis: results.items.filter(a => a.analysisId === 3110)[0],
                        genderAnalysis: results.items.filter(a => a.analysisId === 3111)[0],
                        ageAnalysis: results.items.filter(a => a.analysisId === 3112)[0],
                        participantCountAnalysis: results.items.filter(a => a.analysisId === 3203)[0],
                        versionAnalysis: results.items.filter(a => a.analysisId === 3113)[0]
                        // this.processResults(q, this.survey.participantCount);
                    };
                    questionWithResults.countAnalysis.results.sort((a1, a2) => {
                        if (a1.countValue > a2.countValue) {
                            return -1;
                        }
                        if (a1.countValue < a2.countValue) {
                            return 1;
                        }
                        return 0;
                    });
                    this.setState({questionWithResults: questionWithResults, showAnswers: true});
```

public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,138 @@
...
+    showAnswers(e?) {
+        if (e && e.key !== 'Enter') {
+            return;
+        }
+        this.getAnalyis();
+        setTimeout(() => {
+            this.setState({
+                showAnswers: !this.state.showAnswers
+            });
+        }, 100);
```
dolbeew: We should add a condition so we only call the api once:
```suggestion
        if (!this.state.questionWithResults) {
            this.getAnalyis();
        } else {
            this.setState({
                showAnswers: !this.state.showAnswers
            });
        }
```
I think we can also remove the `setTimeout` if we set `showAnswers` to true after the api returns.

public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    getAnalyis() {
```
dolbeew: Nit: misspelled function name

public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+    showAnswers: boolean;
+    questionWithResults: any;
+}
+
+export class SurveyQuestionReactComponent extends React.Component<Props, State> {
+    constructor(props: Props, state: State) {
+        super(props);
+        this.state = {
+            showAnswers: false,
+            questionWithResults: {}
```
dolbeew: Opt nit: I would usually init a state var like this to `null` or `undefined`, makes the condition simpler when checking if it's set (`!this.state.questionWithResults` vs `!this.state.questionWithResults.countAnalysis`)

public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    showAnswers(e?) {
+        if (e && e.key !== 'Enter') {
+            return;
+        }
+        this.getAnalysis();
+
+        this.setState({
+            showAnswers: !this.state.showAnswers
+        });
```
dolbeew: This will still call the api every time we show/hide the answers, even when the data is already in state. We should check if the data is there before calling:
```suggestion
        if (!this.state.questionWithResults.countAnalysis) {
            // Will need to set showAnswers to true in getAnalysis along with this change
            this.getAnalysis();
        } else {
            this.setState({
                showAnswers: !this.state.showAnswers
            });
        }
```

public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+    searchTerm: string;
+}
+
+interface State {
+    showAnswers: boolean;
+    questionWithResults: any;
+}
+
+export class SurveyQuestionReactComponent extends React.Component<Props, State> {
+    constructor(props: Props, state: State) {
```
dolbeew: No need for a state arg here in the constructor:
```suggestion
    constructor(props: Props) {
```

public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+            showAnswers: false,
+            questionWithResults: null
+        };
+    }
+
+    showAnswers(e?) {
+        if (e && e.key !== 'Enter') {
+            return;
+        }
+        if (!this.state.questionWithResults.countAnalysis) {
```
dolbeew: Since you switch the initial value to null, drop `.countAnalysis` or you'll get an error when trying to show the answers.
```suggestion
        if (!this.state.questionWithResults) {
```

public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+                    questionWithResults.countAnalysis.results.sort((a1, a2) => {
+                        if (a1.countValue > a2.countValue) {
+                            return -1;
+                        }
+                        if (a1.countValue < a2.countValue) {
+                            return 1;
+                        }
+                        return 0;
+                    });
+                    this.setState({ questionWithResults: questionWithResults });
```
dolbeew, line 98: Need to set `showAnswers` here for it to show the first time in:
```suggestion
                    this.setState({ questionWithResults: questionWithResults, showAnswers: true });
```

dolbeew (approved):

public-ui/src/app/data-browser/views/survey-view/components/survey-question-react.component.tsx
```diff
@@ -0,0 +1,199 @@
...
+                console.log(results, 'results');
+                this.setState({ surveyCountAnalysis: results });
+                // if (this.surveyCountAnalysis) {
+                //     localStorage.setItem('surveyCountAnalysis', JSON.stringify(results));
+                // }
+            });
+    }
+
+    getSurveyVersionAnalysis(): any {
+        api.getSurveyVersionCounts(this.props.surveyConceptId)
```
SrushtiGangireddy: This can be fetched from survey-view.component.ts or removed from there may be to avoid multiple fetches.

PR #844, opened 2021-05-03
[risk=no] Add navigation functions for use in React components

dolbeew:
`navigate()` takes an array of strings of the desired route path
`navigateByUrl()` takes a string of the entire url


SrushtiGangireddy (approved):
Thanks. Looks good.

PR #855, opened 2021-05-19
[RW-6757][risk=no] Add fetch swagger api clients

dolbeew:
- Follows the same method as the Workbench for client-side api calls
- Converted all api calls currently being made in React components


SrushtiGangireddy (approved):
Thanks. Looks good to me.

PR #861, opened 2021-06-03
[risk=no] Add stores for route data

dolbeew:
Adds `urlParamsStore`, `queryParamsStore` and `routeConfigDataStore` store for accessing route params and data in React components. `routeConfigDataStore` will most likely only be used for breadcrumbs once we convert it.


SrushtiGangireddy (approved):
Looks good to me. Thanks.

PR #863, opened 2021-06-03
[risk=no] [RW-6776] Converting survey route to react

SrushtiGangireddy:


public-ui/src/app/data-browser/views/survey-view/survey-react-view.component.tsx
```diff
@@ -184,16 +185,18 @@ interface State {
...
     constructor(props) {
         super(props);
+        const {id, search} = urlParamsStore.getValue();
         this.state = {
             isCopeSurvey: false,
             survey: null,
             surveyPdfUrl: '',
-            searchWord: this.props.searchWord,
+            domainId: id ? id : this.props.domainId,
+            searchWord: search ? search : this.props.searchWord,
```
dolbeew, line 199: Seems like these always use the values from the store and the values in props are never set here. Is there any scenario where the props will be used here?

public-ui/src/app/data-browser/databrowser-routing.module.ts
```diff
@@ -93,9 +106,15 @@ const routes: Routes = [
...
   }
-
 ];
 
+if (reactSurvey) {
+    for (const route of routes.filter(r => r.path === '')[0].children[0].children.filter(c => c.path.indexOf('survey') > -1)) {
+        route.data = {};
+        route.component = AppRouting;
+    }
+}
```
dolbeew, line 116: It's probably best if we wait until we're ready to remove the Angular wrapper from the component to convert the route. Then we won't need conditions like this and will also allow us to wrap the React component with `withRouteData` instead of adding a const to `AppRoutingComponent`

public-ui/src/app/app-routing.tsx
```diff
@@ -1,10 +1,13 @@
 import {Component as AComponent} from '@angular/core';
 import * as React from 'react';
 
-import {AppRoute, AppRouter} from 'app/components/app-router';
+import {AppRoute, AppRouter, withRouteData} from 'app/components/app-router';
 import {BaseReactWrapper} from 'app/data-browser/base-react/base-react.wrapper';
+import {SurveyViewReactComponent} from 'app/data-browser/views/survey-view/survey-react-view.component';
 import {IntroVidReactComponent} from './data-browser/views/intro-vids/intro-vids-react.component';
 
+const SurveyViewPage = withRouteData(SurveyViewReactComponent);
```
dolbeew, line 9: As mentioned in the previous comment, we should wrap `withRouteData` around the actual component like we did with `IntroVidReactComponent`

public-ui/src/app/data-browser/views/survey-view/survey-react-view.component.tsx
SrushtiGangireddy: Only when this component is called from angular version of survey-view.

public-ui/src/app/data-browser/databrowser-routing.module.ts
SrushtiGangireddy: I totally agree with waiting till we remove the wrapper component.

public-ui/src/app/data-browser/views/survey-view/survey-react-view.component.tsx
dolbeew: Ok, so we should be able to drop `domainId` and `searchWord` from props and state, just get them directly from the store wherever we need them like you did above:
`const {id, search} = urlParamsStore.getValue();`

public-ui/src/app/data-browser/databrowser-routing.module.ts
dolbeew: 👍

PR #864, opened 2021-06-04
[risk=no] [RW-6838] fixing sources chart button bug

SrushtiGangireddy:
Clicking on source button opens up biological sex chart. Fixing it to instead open up sources chart.


mhigbyflowers (approved):
Thanks for fixing this!

PR #869, opened 2021-07-01
[risk=no] [RW-6906] Convert fitbit view component and route to react

SrushtiGangireddy:
Convert fitbit view component and route to react


mhigbyflowers (approved):
👍

PR #870, opened 2021-07-06
[risk=no] Fitbit tile shows up when no search results

SrushtiGangireddy:
Fix the bug tht shows up fitbit tile even when there are no matching concepts.


mhigbyflowers (approved):
👍

PR #872, opened 2021-07-06
[risk=no] Survey Params conversion

SrushtiGangireddy:
Converting survey params to route from query


dolbeew (approved):
Looks good. Do we need to keep the three sections that are commented out?

SrushtiGangireddy:
@dolbeew Just wanted them to be in there in case we have to shift to using angular component for some reason. Added TODO in there to clean up later.

dolbeew:
Sounds good, thanks!

PR #881, opened 2021-07-21
setting up envars to push

mhigbyflowers:
reactHomeView: false,
  reactChart: true,
  reactSurveyTable: false,
  reactSurvey: true,
  ehrReact: false,
  testReact: false


SrushtiGangireddy (approved):
Looks good thanks.

PR #885, opened 2021-07-26
[risk=no] fixing count position on tiles

SrushtiGangireddy:


mhigbyflowers (approved):
Thanks for taking care of this. I also aligned everything up for the home-view react component

PR #887, opened 2021-07-29
[risk=no] fixing ehr no sources bug

SrushtiGangireddy:
Ehr View breaks on clicking sources chart expand on concepts that do not have source concepts.


mhigbyflowers (approved):
👍 thanks

PR #896, opened 2021-08-20
[risk=no][RW-7104]React Bread Crumbs

mhigbyflowers:


public-ui/src/app/shared/components/breadcrumb/breadcrumb-react.component.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <span className="crumb-container">
+                <a href={environment.researchAllOfUsUrl} style={styles.preCrumb}>Home</a>
+                <span>{'>'}</span>
+                <a onClick={() => NavStore.navigateByUrl('/')} style={domainName ? styles.preCrumb : styles.crumb}>Data Browser</a>
+                {domainName && <React.Fragment><span>{'>'}</span>
+                    <span style={styles.crumb}>{domainName}</span> </React.Fragment>}
+            </span>
+        </React.Fragment>
+    }
+
```
mhigbyflowers: Simple not router solution to bread crumbs

public-ui/src/app/shared/components/breadcrumb/breadcrumb-react.component.tsx
```diff
@@ -0,0 +1,68 @@
...
+const css = `
+    .crumb-container *{
+        margin-right: 0.5em;
+    }
+    .crumb-container {
+        margin-top:1em;
+        display:block;
+    }
+`;
```
dolbeew, line 27: I think we should move this into the reactStyles object and drop the `<React.Fragment>` and `<style>` tags from the template. It's better to have a single style method, even if we duplicate `marginRight: '0.5em'` a few times. Something like the suggested change above should work.

public-ui/src/app/shared/components/breadcrumb/breadcrumb-react.component.tsx
```diff
@@ -0,0 +1,68 @@
...
+import * as React from 'react';
+
+const styles = reactStyles({
+    preCrumb: {
+        color: '#0079b8'
+    },
+    crumb: {
+        color: '#2b266d'
+    }
+
```
dolbeew, line 16:
```suggestion
    preCrumb: {
        color: '#0079b8',
        marginRight: '0.5em'
    },
    crumb: {
        color: '#2b266d',
        marginRight: '0.5em'
    },
    crumbContainer: {
        margin: '1em 0.5em 0 0',
    },
    separator: {
        marginRight: '0.5em'
    }
```

public-ui/src/app/shared/components/breadcrumb/breadcrumb-react.component.tsx
```diff
@@ -0,0 +1,68 @@
...
+    constructor(props: Props) {
+        super(props);
+
+    }
+
+    render() {
+        const { domainName } = this.props;
+        return <React.Fragment>
+            <style>{css}</style>
+            <span className='crumb-container'>
```
dolbeew, line 44: Can change this to a `<div>` and drop the `display: block` from the styles

dolbeew (commented):
I like the simpler approach for the breadcrumb component, the old implementation was probably overkill for this router, but I think we should keep a single instance in the top-level component instead of duplicating in each view. I can try to find a simple way to get the domain/survey from the router so we can keep this approach.

dolbeew:
Made a new branch off of this one and created a draft pr instead of trying to explain a different approach in the comments. #897

PR #918, opened 2021-10-15
[risk=no][RW-7415]adding 'geno' environment Var

mhigbyflowers:
adding 'geno' environment Var


SrushtiGangireddy (approved):
Thanks. Meant to do it in the other PR.

PR #919, opened 2021-10-18
[risk=no] fetching genomics demographic counts data

SrushtiGangireddy:


mhigbyflowers (approved):
looks good 
one thing: let's remove the 'react' from the file name since it will all be in react for the new components

PR #921, opened 2021-10-19
[RW-7418][Risk=no]Genomics-view component

mhigbyflowers:
image


dolbeew:
This doesn't happen every time, but there seems to be a race condition that sometimes results in an error that prevents the main homepage content from rendering. It occurs when `genomicsApi().getParticipantCounts()` hasn't returned yet when the component renders:
Screen Shot 2021-10-20 at 10 16 52 AM

I think putting a condition around the genomics instance of `ResultLinksComponent` (home-view-react line 533) that checks if `genomicInfo` is set will prevent this from happening.

I don't think this was caused by this pr, but it happened the first time I pulled up this branch locally, so I though I'd mention it.

dolbeew (approved):
Left a comment about an error I encountered, rest of the pr looks good.

PR #926, opened 2021-10-25
Component data issue

mhigbyflowers:


public-ui/src/app/data-browser/views/genomic-view/components/genomic-chart.component.tsx
```diff
@@ -1,34 +1,93 @@
...
+        this.state = {
+            options: null,
+            setData: props.data
+        }
+        console.log(this.props.data, 'from', this.props.title);
+    }
+    
+
+    dataToOptions() {
+        this.chartOptions = baseOptions;
```
dolbeew, line 47: I think the data is getting passed back to the `baseOptions` constant, so all three get left with the last set of data. Seems to work if I make sure the reference is broken using `JSON.parse`/`JSON.stringify`:
```suggestion
        const chartOptions = JSON.parse(JSON.stringify(baseOptions));
```

PR #942, opened 2021-11-15
[DB-1034][Risk=no]removing scroll bar

mhigbyflowers:
this should remove the scroll bar that is showing up on window machines. I am not able to test it on my machine and will have to ping Kevin to look at it on test after its merged


SrushtiGangireddy (approved):
LGTM.

PR #950, opened 2021-11-19
Mhf/db 1047

mhigbyflowers:
Restructured where Api are called ( genomic view) to store the data when changing 'tabs'


SrushtiGangireddy (approved):
Looks good to me.

PR #951, opened 2021-11-22
[risk=no] [DB-1043] Add sorting on gene, allele count, allele number, allele frequency

SrushtiGangireddy:
image

### DIFF
--- a/public-api/db-cdr/generate-cdr/make-bq-data.sh
+++ b/public-api/db-cdr/generate-cdr/make-bq-data.sh
@@ -138,7 +138,7 @@ if [ "$SEARCH_VAT" = true ]; then
   WHERE is_canonical_transcript OR transcript is NULL
   ORDER BY vid, row_number),
   genes as (
-     SELECT vid, ARRAY_TO_STRING(array_agg(distinct gene_symbol ignore nulls ORDER BY gene_symbol), ' ') as genes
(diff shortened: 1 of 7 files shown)


mhigbyflowers:
I am not able to get any column to sort except for Variant.
When I click to sort gene, shouldn't the arrow disappear on Variant?
Maybe there should be a small icon button next to the column that indicates that the column can be sorted.

SrushtiGangireddy:
@mhigbyflowers  So, i was using local api. You should be able to use local api to see sorting on gene, allele count and frequency.

SrushtiGangireddy:
> I am not able to get any column to sort except for Variant. When I click to sort gene, shouldn't the arrow disappear on Variant? Maybe there should be a small icon button next to the column that indicates that the column can be sorted.

When we click on gene sort, the arrow on variant id would change back to the original asc arrow (which is. default).

SrushtiGangireddy:
@mhigbyflowers Let me know if you can get the local api running or i can separate out api part into another ticket.

mhigbyflowers:
@SrushtiGangireddy  OK, I am able it get it to work.
I think there should not be an arrow on any columns unless it is being sorted.

SrushtiGangireddy:
> @SrushtiGangireddy OK, I am able it get it to work. I think there should not be an arrow on any columns unless it is being sorted.

We should then probably think about a different icon or way to let user know that column is sortable. i was thinking we could have the arrow on the column and then have it big when the sort is active. Let me knw what you think.

PR #968, opened 2021-12-16
[risk=no] [DB-1057] remove search term when component changes

SrushtiGangireddy:
Remove search term when component changes
Display allele freq to 6 decimals


mhigbyflowers (approved):
looks good, thanks

PR #971, opened 2022-01-03
[risk=no] Basics survey questions branching logic bug fix

SrushtiGangireddy:
Question “Are any of these a closer description to your gender identity?” is missing branching logic from answers “non-binary”, “transgender”. 

Add note to question:

Have you ever served on active duty in the United States Armed forces, either in the regular military or in a National Guard or military reserve unit?

Note: Active duty does not include training for the Reserves or National Guard, but DOES include activation, for example, for the Persian Gulf War5

Disability section

Remove  [HHS.ACA Section 4302] in DB


mhigbyflowers (approved):
👍

PR #984, opened 2022-02-05
[risk=no] Paginator

SrushtiGangireddy:


mhigbyflowers (commented):
They wanted there only to be 10 results per page as default. Is that a backend change?

mhigbyflowers (approved):
👍

PR #989, opened 2022-02-11
[risk=no] [DB-1106] Removing tooltips from ehr and genomics section titles

SrushtiGangireddy:
Removing tooltips from ehr and genomics section titles


mhigbyflowers (approved):
👍

PR #1009, opened 2022-02-17
[DB-1095][risk=no] puts the changes back but without paginator

mhigbyflowers:
puts the changes back but without paginator


SrushtiGangireddy (approved):
LGTM. Thanks.

PR #1017, opened 2022-02-21
fixing row width issue

mhigbyflowers:
fixed issue: constraining the right edge of the results table to not expand beyond the last column


SrushtiGangireddy (approved):
LGTM. THanks.

PR #1033, opened 2022-03-11
[risk=no] changing search behaviour

SrushtiGangireddy:
change default gene search type from CONTAINS to EXACT MATCH
enable CONTAINS search if term is preceded by tilde (~) (ex. use ~BRCA for CONTAINS search)


mhigbyflowers (approved):
Lgtm

PR #1037, opened 2022-03-16
[risk=no] Api for genomic filter option fetch

SrushtiGangireddy:


public-api/src/main/resources/public-api.yaml
```diff
@@ -97,6 +97,25 @@ paths:
...
+          schema:
+            $ref: "#/definitions/VariantResultSizeRequest"
+      responses:
+        200:
+          description: Variant result list size
+          schema:
+            type: integer
+            format: int64
+
+  /v1/genomics/genomic-filter-options:
```
SrushtiGangireddy, line 113: This api call is to fetch unique genomic filter options.

public-api/src/main/resources/public-api.yaml
```diff
@@ -1178,6 +1197,21 @@ definitions:
       - GENO_AGE_ANALYSIS
       - GENO_RACE_ANALYSIS
 
+  VariantResultSizeRequest:
```
SrushtiGangireddy, line 1200: Changed this input to body to also send in filter metadata.

public-api/src/main/resources/public-api.yaml
```diff
@@ -1468,6 +1506,59 @@ definitions:
         items:
           $ref: "#/definitions/Variant"
 
+  GenomicFilterOption:
```
SrushtiGangireddy, line 1509: This is the json structure used to send the filter options back. Checked would be true initially and when user unchecks and sends it back to api call the results would be eliminated.

SrushtiGangireddy: option is the actual gene/ clinical significance/ consequence value. count is the result count for that option. min and max are only used for allele count / allele number / allele frequency.

public-api/src/main/java/org/pmiops/workbench/publicapi/GenomicsController.java
```diff
@@ -55,7 +59,41 @@
             "gvs_sas_ac as sas_allele_count, gvs_sas_an as sas_allele_number, gvs_sas_af as sas_allele_frequency, " +
             "gvs_oth_ac as oth_allele_count, gvs_oth_an as oth_allele_number, gvs_oth_af as oth_allele_frequency, " +
             "gvs_all_ac as total_allele_count, gvs_all_an as total_allele_number, gvs_all_af as total_allele_frequency from ${projectId}.${dataSetId}.wgs_variant";
-
+    private static final String FILTER_OPTION_SQL_TEMPLATE_GENE = "with a as\n" +
```
SrushtiGangireddy, line 62: This query fetches unique gene value, count
consequence value, count
clinical significance value, count
to display options on UI.

PR #1063, opened 2022-05-26
[DB-1093] Genomic filter 

mhigbyflowers:
The Genomic filter still need work but I wanted to push what I have so far to test.

image

### DIFF
--- a/public-api/src/main/java/org/pmiops/workbench/publicapi/GenomicsController.java
+++ b/public-api/src/main/java/org/pmiops/workbench/publicapi/GenomicsController.java
@@ -675,9 +675,6 @@ public class GenomicsController implements GenomicsApiDelegate {
                 }
             }
         }
-        System.out.println("************************ +++++++++ ****************************");
-        System.out.println(finalSql);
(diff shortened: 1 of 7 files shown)


mhigbyflowers:
Specify the sliders are not affecting the search. Im not sure if it is on the front-end or backend. Or if i am missing something.

SrushtiGangireddy:
> Specify the sliders are not affecting the search. Im not sure if it is on the front-end or backend. Or if i am missing something.

I have added the checked filter for allele count, allele number, allele frequency filters on the backend. Setting the checked to true will let the backend know these filters are on.

SrushtiGangireddy (approved):
This looks so good for preliminary. 

One thing I noticed is the once the filter is applied it stays on the screen by default and goes away when clicked on filter button. 
image

Is this intentional ? Should we not close it be default once the filter is applied and open it when the user clicks on the filter button ?

Also, the slider remains at the start (0) for min and maximum value for max even when the filter is active and min is at say 200. 

We can address these in the next prs though.

PR #1143, opened 2022-10-17
[risk=no] Changing Db Driver and connection string

SrushtiGangireddy:


public-api/src/main/java/org/pmiops/workbench/cdr/CdrDbConfig.java
```diff
@@ -59,6 +59,7 @@ public CdrDataSource(CdrVersionDao cdrVersionDao,
       Map<Object, Object> cdrVersionDataSourceMap = new HashMap<>();
       for (DbCdrVersion cdrVersion : cdrVersionDao.findAll()) {
         int slashIndex = originalDbUrl.lastIndexOf('/');
+        // String dbUrl = "jdbc:mysql:///databrowsermaindb?cloudSqlInstance=aou-db-test:us-central1:databrowsermaindb&socketFactory=com.google.cloud.sql.mysql.SocketFactory";
         String dbUrl =
```
yonghaoy, line 70: You need to put CDR DB name inside 
```
jdbc:mysql:///databrowsermaindb?cloudSqlInstance=aou-db-test:us-central1:{db name}&socketFactory=com.google.cloud.sql.mysql.SocketFactory
```
You need to make sure you are not dropping the part after DB, like `&socketFactory=com.google.cloud.sql.mysql.SocketFactory`

yonghaoy (changes requested):

public-api/src/main/java/org/pmiops/workbench/cdr/CdrDbConfig.java
SrushtiGangireddy: Can i just do this for now. Comment out the original part. Put in the test url and test that out.

yonghaoy (commented):
Per: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/docs/jdbc-mysql.md
We may also need to upgrade connector version
https://github.com/all-of-us/data-browser/blob/49ba38a6ab19417dfd3ecbdf28afac1f0b87ec76/public-api/build.gradle#L307
or 
https://github.com/all-of-us/data-browser/blob/95fca15589ebe33f8e07cfbc0c6c3b693ce7bc38/public-api/build.gradle#L296

PR #1144, opened 2022-10-18
[risk=no] Adding sql instance

SrushtiGangireddy:


yonghaoy (changes requested):
I don't think you can put values inside config. Configs are loaded after serve up and running(then load config from DB). 
To make server up, server need to have correct DB setup. 
Then DB config has to go into application property

PR #1217, opened 2023-02-08
[DB-1236][risk=no]homepage stytling

mhigbyflowers:


SrushtiGangireddy (approved):
Looks good. Not sure if the padding below the survey questions is little bit more than rest. Ignore if the same. 
image

PR #1218, opened 2023-02-10
[risk=no][noticket] fixing missing info on tiles

mhigbyflowers:
fixing bug where 'questions available' and alike were missing


SrushtiGangireddy:
On the initial load the medical concepts text is not visible. Once the search term is entered and cleared it appears though. I am thinking its because of typing being true in resultlinkscomponent.
image

SrushtiGangireddy:
Setting typing to true here fixes the text on initial load too. But I am not sure if it breaks something else.

SrushtiGangireddy (approved):
Looks good to me. Thanks.

PR #1298, opened 2023-09-19
[DB-1290][risk=no] fixes and tweaks to genomic infinite scroll:

mhigbyflowers:
fixes and tweaks to genomic infinite scroll:

fixed: scrolling is held up when user reaches bottom of currently displayed rows and needs to wait as more rows need to load

fixed: potential fix = load more rows, earlier, to avoid pauses when scrolling to bottom of table

fixed: bug when filters/sorting applied when user is scrolled down onto subsequent pages

fixed: potential fix = when applying filters/sorting, automatically return to row 1 of page 1

fixed: Filter & Sort button intermittently disappears when new pages/rows are loading


SrushtiGangireddy (approved):
Looks good. Thanks.

PR #1299, opened 2023-09-19
[risk=np] [DB-1283] Fixing Data Browser Concept ID missing 

SrushtiGangireddy:
Fixing missing concept id by fixing the null check query.


mhigbyflowers (approved):
thanks

PR #1301, opened 2023-09-20
[Db-1278][risk:no]Infinite scroll on EHR

mhigbyflowers:


SrushtiGangireddy:
image

image

The count on the top keeps showing some random results number i guess its the number of results fetched but its sometimes > the number of rows available.

mhigbyflowers:
> image
> 
> image
> 
> The count on the top keeps showing some random results number i guess its the number of results fetched but its sometimes > the number of rows available.

I removed the "showing top #" cause we are now showing all results
image

SrushtiGangireddy:
Looks good, overall. Could we add a loader at the bottom when the results are waiting to load ?
image

During search, the top 10 concepts and the concept count is right but the results are not taking the search word into account i think. Its displaying all the results. 
image

SrushtiGangireddy:
I was getting this when i searched with "lab" on labs and measurements. 
image

SrushtiGangireddy:
Also, when i search with "syndrome" in conditions, the count on says 873 concepts but after 50 it does not load anymore. 
image

SrushtiGangireddy (approved):
Looks good to me. Thanks.

PR #1312, opened 2023-10-12
[risk=no] java build upgrade

SrushtiGangireddy:
Upgrading Java build to 11 while letting the runtime be at 8


yonghaoy (approved):
I am not expert on this but this is similar to RW's change, as long as it works and then LGTM!

PR #1344, opened 2023-12-15
[DB-1331][risk=no]Mobile styling - allow users to tap tooltip (i) without clicking card to open in new screen

mhigbyflowers:


SrushtiGangireddy (approved):
LGTM. Thanks.

PR #1346, opened 2024-01-04
[risk=no] [DB-1343] Vat v2 adding variant type column

SrushtiGangireddy:
Adding in variant type column

image

image

image


mhigbyflowers (approved):
Thanks 👍

PR #1362, opened 2024-01-26
[risk=no] Java upgrade 8 -> 11

SrushtiGangireddy:
trying java upgrade


SrushtiGangireddy:
image (1)

public-api/build.gradle
```diff
@@ -434,7 +455,13 @@ test {
     maxHeapSize = '2048m'
 }
 
+def cloudSdkHomePath = System.getenv('CIRCLECI') == 'true' ?  "/home/circleci/google-cloud-sdk" : "/google-cloud-sdk"
+
+
 appengine {  // App Engine tasks configuration
+    tools {
+        cloudSdkHome = cloudSdkHomePath
+    }
```
jmthibault79, line 464: RWB has a line here `cloudSdkVersion = '392.0.0'` - not sure if that is helpful for you as well

jmthibault79 (approved):
I wouldn't say that I did a very thorough review, but I didn't see anything concerning

public-api/src/main/webapp/WEB-INF/appengine-web.xml.template
```diff
@@ -20,7 +35,7 @@
   </system-properties>
 
   <automatic-scaling>
-    <min-idle-instances>${GAE_MIN_IDLE_INSTANCES}</min-idle-instances>
-    <max-instances>${GAE_MAX_INSTANCES}</max-instances>
+    <min-idle-instances>1</min-idle-instances>
+    <max-instances>10</max-instances>
```
yonghaoy: If those numbers come from RW. I don't know the traffic for data browser so the same number might not be the best for data brower. Worth to check GAE metrics to get what is the machine, instance count you need.

yonghaoy (approved):
as long as it works in GCP, no other questions

public-api/build.gradle
SrushtiGangireddy: I am not sure this line worked for me for some reason. The difference i would assume is because RW was running api without docker so having cloudsdk on local and this version would suffice may be. For me without this path the docker was not somehow able to find the path of gcloud.

PR #1393, opened 2024-03-26
[risk=no] Adding lou gehrig missing question

SrushtiGangireddy:
1. Adding lou gehrig missing question
2. Removing nesting for questions like 'are you still seeing ... ', 'about how old were you when ...', 'are you currently prescribed ...' 
3. Updating test cdr version with these changes.


mhigbyflowers (approved):
👍

PR #1419, opened 2024-05-07
[risk=no] Springupgrade 06 04 2024

SrushtiGangireddy:


public-api/src/main/java/org/pmiops/workbench/interceptors/CorsInterceptor.java
```diff
@@ -1,18 +1,18 @@
 package org.pmiops.workbench.interceptors;
```
yonghaoy: this file is nolonger needed. We might need to delete this

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 
 defaults: &defaults
   docker:
-    - image: allofustest/workbench:aou-db-image-8
+    - image: us.gcr.io/broad-dsp-gcr-public/workbench:buildimage-0.0.28
```
yonghaoy, line 5: will you use this image?

public-api/src/main/java/org/pmiops/workbench/RootController.java
```diff
@@ -8,6 +8,7 @@ public class RootController {
 
   @RequestMapping("/")
   public String index() {
+    System.out.println("~~~~~~~!!!!!RootController");
```
yonghaoy: remove

yonghaoy (approved):
Congrats! 
LG as long as it works in local.

.circleci/config.yml
SrushtiGangireddy: I want to use this image till we get it to test and then will try to generate our own and see if it works.

PR #1437, opened 2024-06-18
[risk=no] trying to fix openapi generator error

SrushtiGangireddy:


public-ui/libproject/devstart.rb
```diff
@@ -54,12 +54,13 @@ def dev_up(*args)
     common = Common.new
     options = BuildOptions.new.parse("dev-up", args)
 
-    install_dependencies
+    install_dependencies()
```
dolbeew, line 57: Nit: Can probably remove this line since you added the lines below. Shouldn't hurt anything to keep it, would just run `yarn install` twice.

dolbeew (approved):

PR #1537, opened 2025-08-19
[risk=no] fixing some location map bugs

SrushtiGangireddy:
1. In participant demographics, the territories fill color is correct for each genomic type, but the states all show up in the short-read WGS gray-green color even when the other 3 genomic types are selected
2. all fitbit tabs are showing the exact same counts, but on prod there are slightly different counts for each one


SrushtiGangireddy:
image
image
image

dolbeew (approved):

